PR #9, opened 2015-09-17
Bug 29133 - Maintain coded group continuity in sequence append mode w…

wolenetz:
…hen resetting parser state.


wolenetz:
@jdsmith3000 please take a look. Thanks!

jdsmith3000:
We've reviewed the PR and agree with it.

PR #32, opened 2015-10-16
Fix issue 30: Removed unnecessary segment parser loop step referencing undefined media segment header

wolenetz:
Jerry, please take a look.

### DIFF
--- a/index.html
+++ b/index.html
@@ -433,7 +433,7 @@ table.simple {
   </p>
   <h1 class="title p-name" id="title" property="dcterms:title">Media Source Extensions</h1>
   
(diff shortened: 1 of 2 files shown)


wolenetz:
(Please don't use the GH "Merge pull request" green button. If you approve of this PR, I'll do the necessary steps to keep the git history free of the additional merge commit as described in @ddorwin's recent p-h-m thread.)

wolenetz:
Jerry- friendly ping :)

jdsmith3000:
This change means that "need more data" will be driven by lack of a complete media segment, and the incomplete media header requirement that was previously in the spec won't be explicitly defined.

PR #33, opened 2015-10-19
Create README.md + labeling and milestone guidance

wolenetz:


wolenetz:
Jerry and Mark, please take a look. I'll send to public-html-media shortly with a reference to this. Essentially, this guidance is my understanding of what Jerry and I did when triaging MSE github issues last week, with a few clarifications and a couple corrections (which I'll also highlight in the p-h-m email I'm about to send...) Thanks!

wolenetz:
As part of this summarization, two action items were discovered. I would like to make sure we're all on the same page for MSE and use this pull request to make sure Mark, Jerry and I agree with the guidance and the following:
1) Generally, we should consider closing V.Next MSE issues (for now, at least).
2) We should change the currently-labeled “needs author input” to “needs feedback” and assign to appropriate reporter to obtain that feedback. During triage, we misunderstood “author” to mean bug author, not web author. We’ll also need to add a separate, new, “needs author input” label for future usage for when we might need input from web authors.

wolenetz:
@jdsmith3000 and @mwatson2, it would be good to get this reviewed, updated if necessary, and landed as soon as possible. Friendly ping :)

README.md
```diff
@@ -0,0 +1,52 @@
...
+Issues with these labels should always be closed (unless they were re-opened at which time an editor should probably remove these labels if the re-opening is accepted).
+
+
+## Milestones
+
+### “V#”:
+A bug in version # of the spec.
+Currently, V1 is the first version of MSE (in CR in Q4 2015).
+
+### “V.Next”:
```
ddorwin: `VNext` probably makes more sense since that will make it alphabetically after V# (i.e. in the Milestones dropdown).

Alternatively, use a little 'v' in all cases: v1, vNext

wolenetz:
I like the alternative (little 'v' in all cases). I'll make that change soon to this PR.

wolenetz:
We also just added a "breaking change" label to track issues that might possibly cause re-entry to CR, for example. Needs to be added to this PR, too.

jdsmith3000:
This draft looks good, Matt.  I agree with using VNext and adding a breaking change description.  I suggest also adding our recent V1Nonbreaking milestone.  With those changes, we should accept this PR.

jdsmith3000:
To confirm, I think we should match the EME milestones, using capital "V" and no decimals.  They are currently "V1", "V1NonBlocking" and "VNext".  I plan to edit these in a triage pass today.

wolenetz:
I've updated the PR with a rebase and to add "V#NonBlocking" milestone, "breaking change" label, and rename "V.Next" to "VNext". I believe this satisfies all comments on the original PR and I'll merge this PR soon.

Issue #14, opened 2015-10-13
Update SourceBuffer.appendStream() and related algorithms to use ReadableByteStream


wolenetz:
Migrated from w3c bugzilla tracker. For history prior to migration, please see:
https://www.w3.org/Bugs/Public/show_bug.cgi?id=27239.

wolenetz:
Needs follow-up by editors to confirm whether this is blocked still by churning normative dependency on Streams API. We could also consider marking this as an at-risk feature in the spec.

wolenetz:
From discussion with Paul Cotton, we need a stable Streams spec at minimum (even if not a w3c normative reference, and just a whatwg reference.)

wolenetz:
I discussed the concerns about stable Streams API spec reference with @domenic and received the following advice:
1) [Streams API spec editors] will never  break incoming anchor links without first coordinating with all consumer specs so they can update.
2) Commit snapshots are available at https://streams.spec.whatwg.org/commit-snapshots/?C=M;O=D if MSE desires the ability to link to a historical reference to what [STREAMS] looked like at a particular point in time, e.g. in something like “NOTE: the algorithms here are based on the state of [STREAMS] as of [its b98007c… commit snapshot.”

See also Paul Cotton's earlier comment in the w3c bug at https://www.w3.org/Bugs/Public/show_bug.cgi?id=27239#c8 that seemed to indicate that ReadableByteStream had a specific reference available for use in MSE spec: https://streams.spec.whatwg.org/#rbs

wolenetz:
(Pending Jerry's update from TPAC MSE f2f) w.r.t "MattWolenetz: just want jdsmith to respond to github proposal saying "concern would be satisfied w/ stable reference"
[17:56] <timeless> jdsmith: that's fine"

jdsmith3000:
If we can define a stable implementation based on the new streams api, then I would support making that change.  It aligns with previous commitments made by the task force on this topic.

yfuna:
@wolenetz, @jdsmith3000  : [The current spec for ReadableByteStream](https://streams.spec.whatwg.org/#rbs) still looks immature, lacking most of its algorithm definitions. I'd like to make sure with you that making it okay to allow a normative reference to such a spec is in line with or a part of the TF's decision in the past.

wolenetz:
[...]

PR #38, opened 2015-11-10
Streams API changed its IDs

plehegar:
This fixes references into the WHATWG stream API

### DIFF
--- a/media-source.js
+++ b/media-source.js
@@ -371,13 +371,13 @@
 
     'performance-now': { func: hrtime_helper, fragment: 'dom-performance-now', link_text: 'Performance.now()',  },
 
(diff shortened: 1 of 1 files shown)


wolenetz:
Many of the links resulting from this PR become links into WriteableStream. Until issue #14 is fixed, the current state of the appendStream() spec (and its links into the STREAMS API spec) is incorrect. I suggest we don't attempt to fix these stale links until #14 is fixed, since a change like this PR might give the incorrect appearance that the appendStream() API is meant to really take in a WriteableStream.

jdsmith3000:
Marked this blocked since it's waiting on resolution if issue #14.

PR #39, opened 2015-11-16
Added extra headers for repo, mailing list, and implementation status

plehegar:


plehegar:
Merging since it's supposed to be harmless

Issue #1, opened 2015-01-08
Change HTTP Referer


carbos10:
My player use Media Source Extensions, i would change the Headers (Referer)  to a external video , how i can do it? ( i've asked and people tell me, i can change the Headers only with MSE)

heff:
The only reason MSE would make that possible is because using MSE requires you to make the XMLHttpRequest to fetch the video file data yourself, and so you could change the headers of your own XMLHttpRequest. However if you're just trying to play back an MP4 file and change the headers, using MSE would be overkill and would take a ton of work.

If the server that hosts the video file you're trying to play is blocking referrers other than their own domain, then they probably don't want you playing their content on another site and you shouldn't be trying to. Otherwise if you have a legitimate reason for accessing the content, your best be would be to download the video and host it on a sever you control.

(I think this issue can be closed)

carbos10:
i can't download video, because i don't have enough space...
I've tried setHeaderRequest when i do the XMLHttpRequest, but it doesn't change and the creators of the webplayer tell me for do it, i must see the Media Source Extensions...
So, there are method for change header requestion?

acolwell:
Media Source Extensions is not involved with fetching the video data. That is up to the web application and is usually achieved via XMLHttpRequest as @heff says. Changing certain headers in an XMLHttpRequest is controlled by CORS. Look at the XMLHttpRequest & CORS specs to figure out whether they will let you change the Referer. My guess is that they won't, but you should double check.

acolwell:
Closing this issue since it isn't a problem with the MSE spec.

PR #42, opened 2015-11-22
Fix #5: specifying seekable for live media

### LINKED ISSUE #5: Seekable differs from non-MSE behavior
The MSE spec seems to indicate that the highest end time for seekable should [never exceed the highest buffered time](https://w3c.github.io/media-source/#h-htmlmediaelement-extensions) when the duration is set to Infinity. The HTML standard indicates that [duration should be Infinity](https://html.spec.whatwg.org/multipage/embedded-content.html#dom-media-duration) for unbounded or live media and that user agents [should be very liberal determining seekable ranges](https://html.spec.whatwg.org/multipage/embedded-content.html#dom-media-seekable) for media.

Safari on iOS and OSX seems to have interpreted this as meaning that the seekable range should include the time ranges covered by all segments in the current "sliding window" of content in a live HLS video. That definition is convenient because it makes seeking to the live point or building a DVR interface a simple operation for downstream developers, and seems in keeping with the spirit of the HTML standard. It does not seem possible to configure Source Buffers or a Media Source to achieve the same effect. Is there a mechanism to override seekable with out-of-band info like you might get from an M3U8?

dmlap:
Add a new method to the MediaSource object that allows additional seekable TimeRanges to be specified for media with a duration of positive Infinity. This allows HTMLMediaElement behavior when using Media Source Extensions with live content to be less restrictive and more consistent with standard media playback.

### DIFF
--- a/media-source-respec.html
+++ b/media-source-respec.html
@@ -520,6 +520,19 @@
           </ol>
         </dd>
 
+        <dt>void addSeekableRange(double start, double end)</dt>
+        <dd>
+          <p>Adds a new time range to the media element's <a def-id="videoref" name="dom-media-seekable">seekable</a> ranges.</p>
+          <ol class="method-algorithm">
+            <li>If <a def-id="duration"></a> is not positive Infinity, then throw an <a def-id="invalid-state-error"></a> exception and abort these steps.</li>
+            <li>If <var>start</var> or <var>end</var> is negative or NaN, then throw an <a def-id="invalid-access-error"></a> exception and abort these steps.</li>
+            <li>If <var>start</var> is not strictly less than <var>end</var>, then throw an <a def-id="invalid-access-error"></a> exception and abort these steps.</li>
+            <li>
+              If <var>seekableUpdates</var> is unset, let it be a new <a def-id="timeranges"></a> consisting of a single range with a start time of <var>start</var> and end time equal to <var>end</var>. Otherwise, let the new value of <var>seekableUpdates</var> be the <a def-id="normalized-timeranges-object"></a> formed by adding the single time range between <var>start</var> and <var>end</var> to its current value.
+            </li>
+          </ol>
+        </dd>
+
         <dt>void endOfStream(optional EndOfStreamError error)</dt>
         <dd>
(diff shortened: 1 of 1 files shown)


dmlap:
I wasn't present at the f2f discussion so sorry in advance if this PR doesn't take into account any additional complexities brought up at that group. It seemed like there was reasonable consensus in the issue so I thought I'd take a stab at a proposal. Comment on the implementation questions from #5:

> Implementation proposal will likely need to consider:
> 1) disallow setting seekableRange when updating is true (app should first let append/remove updating finish, or call abort to make it (append) finish).
> 2) disallow setting seekableRange that would clip any buffered media (this is meant to prevent weird edge cases where there's buffered media in the timeline that can't be seeked to)
> 3) perhaps allow setting empty seekable range (if #1 is satisfied, but not necessarily #2), to allow apps to prevent any seeking. (If the loop flag is set, would this prevent looping back to time 0, or is that an artifact just in Chrome?)

This proposal follows the path implied by 2) which seems to circumvent the need to disallow modifying `seekable` during updates noted in 1). I couldn't think of a good use-case for making `seekable` a strict subset of `buffered` so I didn't attempt to implement that.

jdsmith3000:
I believe the changes here are complete, and reflect the general agreement from TPAC and my request that the seekable ranges additions apply only if duration was positive infinity.  The names are `addSeekableRange` and `removeSeekableRange`, which seems reasonable to me now.

We should add a change record note as well.

dmlap:
@jdsmith3000 I could add the change record but I couldn't add an accurate change date since it's not merged. Was that a note for yourself?

jdsmith3000:
Change record is a note for me.  Matt:  Do you agree this change is ready to apply?  If so, I'll update the change record and approve the PR.

wolenetz:
@jdsmith3000 Please see email thread from chair to us on Dec 7, 2015 and my reply on Dec 8, 2015. We cannot accept this PR at the moment, IIUC.

wolenetz:
Closing this. To be replaced by a new PR generated by me.

Issue #34, opened 2015-10-20
InvalidAccessError usage is questionable


ddorwin:
Most/all uses of `InvalidAccessError` should be replaced with `TypeError` or other values. See w3c/encrypted-media#103 for more information.

jdsmith3000:
This is a non-controversial change to use `TypeError` where it's appropriate.  Explanatory language from the referenced EME issue is:

TypeError is defined in https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror. It is the "default" error type, for cases when something goes wrong and you don't expect authors to do any switching logic on  err.name . It's not just used for type-system errors as you might expect. Invalid values of any type generally fit under TypeError (unless you really want to call out a specific facet of their invalid-ness for authors to switch on and recover from). IIUC, so would data that cannot be sanitized.

wolenetz:
This might be considered a breaking change.

jdsmith3000:
It's our general view that breaking changes are best done now, rather than after a v1 spec is finished.  

I see 12 `InvalidAccessError `instances in the spec.  I'll prepare a change to move them to `TypeError`.

jdsmith3000:
@Wolenetz I've got commits on two issues under PR #47 because I didn't use a branch to create them.  I think it is simplest to leave them combined for these changes.  I'll use branches for editing MSE going forward.

jdsmith3000:
PR #47 resolved this issue in addition to issue #23.

PR #47, opened 2016-01-30
Fix #23 & #34

### LINKED ISSUE #23: Coded frame removal algorithm may introduce discontinuity at end of current coded frame group
https://www.w3.org/Bugs/Public/show_bug.cgi?id=28557

jdsmith3000:
Inserted random access point steps when the frame being removed has a
decode timestamp == the last decode timestamp for that frame's track.


ddorwin:
If this fixes the issue, s/Issue/Fix/ in the description.

jdsmith3000:
Done.

wolenetz:
(LGTM)

Issue #18, opened 2015-10-13
Should fetch algorithm failure trigger detaching from a media element?


wolenetz:
Migrated from w3c bugzilla tracker. For history prior to migration, please see:
https://www.w3.org/Bugs/Public/show_bug.cgi?id=27854

It was previously assigned to Adrian Bateman. Editors will sync soon to determine who to take this bug.

jdsmith3000:
The resource selection algorithm will ultimately set the networkState to `NETWORK_EMPTY` if the media element has neither a src attribute or a source element child.  This seems like a sufficient trigger for the MSE algorithms.

foolip:
Yes, but another possibility is that there is a src attribute or source element child, in which cast networkState will not reach `NETWORK_EMPTY`. Instead, the media element will be playing a new resource, while the old `MediaSource` object is still attached. (That this doesn't happen in implementations is because they don't match the spec.)

wolenetz:
Thanks @foolip (see also #17). We'll likely have the MSE spec define an algorithm that the HTML spec calls, so that in short term we don't block MSE on HTML spec change (both can and should change).

jdsmith3000:
Same comment here as for issue #17.  Can the existing "**Detaching from a media element**" algorithm be made applicable to clarifying that it is also initiated when networkState reaches NETWORK_NO_SOURCE?

foolip:
Tying it to `networkState` reaching `NETWORK_NO_SOURCE` is not ideal, because that isn't the "reset hook" that other things in HTML uses (the media element load algorithm is) and that's also not how it's implemented in at least Blink.

(Note that `networkState` is set to `NETWORK_NO_SOURCE` as the first step of the resource selection algorithm, but the HTML spec is written such that things are already reset when you reach that point.)

jdsmith3000:
Both Issues #17 and #18 should resolve with pull request #56.

foolip:
I think that this issue should remain open, because there isn't yet a call in HTML to the detach steps for the error cases, only for the load algorithm (#17).

wolenetz:
Re-opening per https://github.com/w3c/media-source/issues/18#issuecomment-211395205. @jdsmith3000 please take a look - can this be resolved similarly by non-normative note(s) in MSE? Are additional call-out points in the informative HTML51 reference needed (and if so, that informative reference snapshot URL in MSE spec will also need updating).

jdsmith3000:
[...]

PR #56, opened 2016-03-11
Fixed #17 & #18: Convert "Detach from Media Element" into an externally referenced algorithm

### LINKED ISSUE #17: Should the media element load algorithm trigger detaching from a media element?
https://www.w3.org/Bugs/Public/show_bug.cgi?id=27843

It was previously assigned to Adrian Bateman. Editors will sync soon to determine who to take this bug.

jdsmith3000:
Generalized the MediaSource detach algorithm so that it can be called
for other HTMLMediaElement conditions.

resource=#mediasource-detach

### DIFF
--- a/index.html
+++ b/index.html
@@ -1126,8 +1126,9 @@ interface <span class="idlInterfaceID">MediaSource</span> : <span class="idlSupe
 
         <section id="mediasource-detach" typeof="bibo:Chapter" resource="#mediasource-detach" property="bibo:hasPart">
           <h4 id="h-mediasource-detach" resource="#h-mediasource-detach"><span property="xhv:role" resource="xhv:heading"><span class="secno">2.4.2 </span>Detaching from a media element</span></h4>
-          <p>The following steps are run in any case where the attached <a href="#idl-def-MediaSource" class="idlType"><code>MediaSource</code></a>, if any, must be detached from the media element.
-          </p><ol>
+          <p>The following steps are run in any case where the media element is going to transition to <a href="https://www.w3.org/TR/html5/embedded-content-0.html#dom-media-network_empty">NETWORK_EMPTY</a> and <a href="https://www.w3.org/TR/html5/webappapis.html#queue-a-task">queue a task</a> to <a href="https://www.w3.org/TR/html5/webappapis.html#fire-a-simple-event">fire a simple event</a> named <a href="https://www.w3.org/TR/html5/embedded-content-0.html#event-mediacontroller-emptied">emptied</a> at the media element. These steps must be run right before the transition.</p>
+
+          <ol>
             <li>Set the <code><a href="#widl-MediaSource-readyState">readyState</a></code> attribute to <code><a href="#idl-def-ReadyState.closed">"closed"</a></code>.</li>
             <li>Update <code><a href="#widl-MediaSource-duration">duration</a></code> to NaN.</li>
             <li>Remove all the <a href="#idl-def-SourceBuffer" class="idlType"><code>SourceBuffer</code></a> objects from <code><a href="#widl-MediaSource-activeSourceBuffers">activeSourceBuffers</a></code>.</li>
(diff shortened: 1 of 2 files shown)


foolip:
I have prepared https://github.com/whatwg/html/pull/876 as the companion to this change. It would be good to merge them at the same time, so I guess first let's wait until both PRs are reviewed. Feedback appreciated!

foolip:
OK, the HTML side is merged, if this is merged we can move on to the question of improvements, to match implementations.

paulbrucecotton:
Can @jdsmith3000 please open a bug on the W3C HTML 5.1 spec to ensure the HTML change is made in W3C HTML 5.1?  MSE does not to the WHATWG version of HTML so the WHAT WG change does not help us here.

jdsmith3000:
@paulbrucecotton:  [W3C HTML Issue #155](https://github.com/w3c/html/issues/155) entered.

foolip:
OK to merge this now?

jdsmith3000:
We are awaiting update of W3C/HTML so that this change is compatible ([W3C HTML PR #157](https://github.com/w3c/html/pull/157)).

paulbrucecotton:
@jdsmith3000, @yfuna :  If MSE is going to refer to a change that is only in HTML5.1, should we also be changing the MSE normative reference to HTML since it current references HTML5 not HTML5.1?  Will having a normative reference to HTML5.1 be a problem to getting MSE to Recommendation ASAP? 

See: https://www.w3.org/TR/media-source/#normative-references
[HTML5] Ian Hickson; Robin Berjon; Steve Faulkner; Travis Leithead; Erika Doyle Navara; Edward O'Connor; Silvia Pfeiffer. HTML5. 28 October 2014. W3C Recommendation. URL: http://www.w3.org/TR/html5/

jdsmith3000:
If we keep the current change, then the normative reference will have to change to complete the scenario.  It doesn't seem advisable either to maintain the previous NETWORK_EMPTY condition for running it, while at the same time exposing an algorithm to HTML for more precise control.

@paulbrucecotton Who can best advise on the impact of changing this reference on getting the Recommendation?  If the impact is significant, might we opt to decline the PR and mark the two issues VNext?

yfuna:
@paulbrucecotton, @jdsmith3000 I'm currently creating a straw evaluation of the impact, which would give us a rough image and how we can evaluate the impact more precisely.

I'm also checking the risk from the PoV of the difference in the schedules of MSE and HTML5.1.

yfuna:
I confirmed that the difference in the schedules (HTML5.1 to become a CR when MSE to become a PR) isn't a problem with regard to the MSE PR transition request. So, schedule-wise, it's okay to refer HTML5.1.

jdsmith3000:
@wolenetz Please review.  I believe this change is ready to merge.  It was waiting only on an HTML5.1 reference, and that is now included.

wolenetz:
If I read this correctly, MSE spec takes a normative reference on HTML5, but not HTML5.1. It has a non-normative note that indicates where HTML5.1's new call-out for detachment should trigger the MSE logic. This means that all the other text in the MSE spec that relates to specific extensions to HTML5.x refers normatively to the HTML5, not HTML5.1 spec (and eases this change). Is this the intent, @yfuna or @paulbrucecotton ? If so, I don't disagree, but it wasn't obvious to me from this pull request's comment thread (in fact, I read https://github.com/w3c/media-source/pull/56#issuecomment-200885738 as implying differently that the normative reference should change to HTML5.1).

jdsmith3000:
The [[HTML51]] notation adds the current informative reference.  We might move the link for the non-normative note to normative text, if that would help.

wolenetz:
I see that. My question of @yfuna or @paulbrucecotton is whether or not the normative html5 reference was expected to need to change to html5.1 for all of this mse spec as part of this PR, or if using an informative reference for additional clarity (from a non-normative note as in this current PR, or even from normative text possibly) is ok instead. Particularly, since HTML5 https://www.w3.org/TR/html5/embedded-content-0.html#media-elements makes no mention at all of "detach", the normative portion changed in this PR is now less clear.

jdsmith3000:
I considered including the previous spec language to initiate detach, but it's effectively superseded by the calling condition in HTML5.1, and so doing that is undesirable.  I'm not sure how a normative reference would be made to HTML 5.1 if we wanted to do one, since we'd be referencing a working draft.

wolenetz:
If we have a normative reference on HTML5, we need to clearly describe when MediaSource detach should occur relative to HTML5. We can also include a non-normative note like you have that mentions for further clarity how HTML5.1 calls out precisely when MediaSource detach should occur. As this PR currently is, we have a normative reference on HTML5, but less clearly describe when the detach should occur relative to HTML5 than prior to this PR IMHO.

jdsmith3000:
I will take a cut at adding a normative statement for HTML5, and will likely move the 5.1 reference as well.

jdsmith3000:
Also closes issue #18.

wolenetz:
@jdsmith3000 and I synced externally on a snapshot of his approach in 3cf0e6e and for posterity, this looks good to me, and doesn't introduce risk of MSE v1 spec delay that might occur if we had to transition all normative HTML5 references from this spec to HTML5.1 references.

Issue #26, opened 2015-10-13
Per spec, no frame will be removed when changing mediasource duration


wolenetz:
Migrated from w3c bugzilla tracker. For history prior to migration, please see:
https://www.w3.org/Bugs/Public/show_bug.cgi?id=28727

wolenetz:
It looks like there is strong overlap with some of the other abort on duration reduction issue(s) that Jerry is working on right now.

jdsmith3000:
We intend to separate range removal operations from duration reductions, and that should resolve the primary concern in https://www.w3.org/Bugs/Public/show_bug.cgi?id=28727.

jdsmith3000:
@wolenetz and I had a private discussion on this bug.  This comment summarizes the discussion.

Comments apply to issues #19, #20 and #26.  The issues generally are:
 
- Shortening duration doesn’t work properly with range removal
- `abort()` similarly doesn’t give deterministic results during removes
- Completing range removals async can lead to duration change errors
 
The async issue is a result of a change sometime back to have Coded Frame Removals complete asynchronously when called from Range Removal, with control returned to the caller.
 
A general proposal I previously reviewed with @wolenetz was:
 
-  Require running `remove()` to run Range Removal (by removing automatic removal from Duration Change)
  - From @wolenetz: More precisely, change the spec so only remove() runs range removal algorithm. (Remove the conditional async race problem from the duration change algorithm.) Note that coded frame removal algorithm (which is not the larger range removal algorithm) will still be used by coded frame eviction algorithm as part of append processing.
- Insert a note in Duration Change about not performing automatic range removal (i.e. `remove()` required first)
  - From @wolenetz: Two bullets down from this covers this bullet better, right? (Throw InvalidStateError !! prior to step 3 "Update duration to new duration" !!, if "new duration" < the highest end time reported by the buffered attribute across all SourceBuffer objects in SourceBuffers.
- Duration change would still check for a reduced duration, and possibly throw an error if `updating `is true (i.e. a `remove()` is in process)
  - From @wolenetz: Don't introduce this redundancy. In fact, we run duration change algorithm *to increase duration* as part of coded frame processing, which is part of append processing, which indeed occurs while updating is true. We already have the MediaSource.duration InvalidStateError gate in the attribute setter.
- Throw an `InvalidStateError `exception on Duration Changes that require range removal (revised step 4)
[...]

Issue #20, opened 2015-10-13
Clarify asynchronicity in duration change algorithm when reducing duration


wolenetz:
Migrated from w3c bugzilla tracker. For history prior to migration, please see:
https://www.w3.org/Bugs/Public/show_bug.cgi?id=27982

It was previously assigned to Adrian Bateman. Editors will sync soon to determine who to take this bug.

jdsmith3000:
We plan to disallow aborts of a range removal to insure results are deterministic, and separate range removal from duration changes that would currently do an automatic remove. That affects both this issue and [Issue 19](https://github.com/w3c/media-source/issues/19).

jdsmith3000:
See latest comment in issue #19.  We want to return an error on abort() during range removal.

wolenetz:
@jdsmith3000 fixed this with https://github.com/w3c/media-source/commit/b386c1ca8637e8370f5817e35d71cb68d8854425 merged in https://github.com/w3c/media-source/pull/65

PR #65, opened 2016-04-19
Fixed #19, #20, #26: Require remove() for range removal

### LINKED ISSUE #19: Ambiguous behaviour when running sourcebuffer.abort()
https://www.w3.org/Bugs/Public/show_bug.cgi?id=27980

jdsmith3000:
- Require remove() for all Range Removals
- Error on Duration Changes that need remove first
- Error on abort() during Range Removals

### DIFF
--- a/index.html
+++ b/index.html
@@ -454,7 +454,7 @@ table.simple {
   </p>
   <h1 class="title p-name" id="title" property="dcterms:title">Media Source Extensions</h1>
   
-  <h2 id="w3c-editor-s-draft-03-may-2016"><abbr title="World Wide Web Consortium">W3C</abbr> Editor's Draft <time property="dcterms:issued" class="dt-published" datetime="2016-05-03">03 May 2016</time></h2>
+  <h2 id="w3c-editor-s-draft-12-may-2016"><abbr title="World Wide Web Consortium">W3C</abbr> Editor's Draft <time property="dcterms:issued" class="dt-published" datetime="2016-05-12">12 May 2016</time></h2>
   <dl>
     
       <dt>This version:</dt>
@@ -1327,19 +1327,16 @@ interface <span class="idlInterfaceID">MediaSource</span> : <span class="idlSupe
           <p>Follow these steps when <code><a href="#widl-MediaSource-duration">duration</a></code> needs to change to a <var>new duration</var>.</p>
           <ol>
             <li>If the current value of <code><a href="#widl-MediaSource-duration">duration</a></code> is equal to <var>new duration</var>, then return.</li>
-            <li>Set <var>old duration</var> to the current value of <code><a href="#widl-MediaSource-duration">duration</a></code>.
-            </li><li>Update <code><a href="#widl-MediaSource-duration">duration</a></code> to <var>new duration</var>.</li>
-            <li>If the <var>new duration</var> is less than <var>old duration</var>, then run the
-              <a href="#sourcebuffer-range-removal">range removal</a> algorithm with <var>new duration</var> and
-              <var>old duration</var> as the start and end of the removal range.
-
-              <div class="note"><div class="note-title marker" aria-level="5" role="heading" id="h-note15"><span>Note</span></div><p class="">This preserves audio frames and text cues that start before and end after the <code><a href="#widl-MediaSource-duration">duration</a></code>.</p></div></li>
-            <li>If a user agent is unable to partially render audio frames or text cues that start before and end after the <code><a href="#widl-MediaSource-duration">duration</a></code>, then run the following steps:
+			<li>If <var>new duration</var> is less than the highest starting <a href="#presentation-timestamp">presentation timestamp</a> of any buffered <a href="#coded-frame">coded frames</a> for all <a href="#idl-def-SourceBuffer" class="idlType"><code>SourceBuffer</code></a> objects in <code><a href="#widl-MediaSource-sourceBuffers">sourceBuffers</a></code>, then throw an <code><a href="https://www.w3.org/TR/html5/infrastructure.html#invalidstateerror">InvalidStateError</a></code> exception and abort these steps.</li>
+			  <div class="note"><div class="note-title marker" aria-level="5" role="heading" id="h-note15"><span>Note</span></div><p class="">Duration reductions that would truncate currently buffered media are disallowed. When truncation is necessary, use <code><a href="#widl-SourceBuffer-remove-void-double-start-unrestricted-double-end">remove()</a></code> to reduce the buffered range before updating <code><a href="#widl-MediaSource-duration">duration</a></code>.</p></div>
+            <li>Set <var>old duration</var> to the current value of <code><a href="#widl-MediaSource-duration">duration</a></code>.</li>
+            <li>Update <code><a href="#widl-MediaSource-duration">duration</a></code> to <var>new duration</var>.</li>
+            <li>If a user agent is unable to partially render audio frames or text cues that start before and end after the <code><a href="#widl-MediaSource-duration">duration</a></code>, then run the following steps:</li>
+			  <div class="note"><div class="note-title marker" aria-level="5" role="heading" id="h-note16"><span>Note</span></div><p class="">This condition can occur because the <a href="#sourcebuffer-coded-frame-removal">coded frame removal algorithm</a> preserves audio frames and text cues that start before and end after the <code><a href="#widl-MediaSource-duration">duration</a></code>.</p></div>
               <ol>
                 <li>Update <var>new duration</var> to the highest end time reported by the <code><a href="#widl-SourceBuffer-buffered">buffered</a></code> attribute across all <a href="#idl-def-SourceBuffer" class="idlType"><code>SourceBuffer</code></a> objects in <code><a href="#widl-MediaSource-sourceBuffers">sourceBuffers</a></code>.</li>
                 <li>Update <code><a href="#widl-MediaSource-duration">duration</a></code> to <var>new duration</var>.</li>
               </ol>
-            </li>
             <li>Update the <code><a href="https://www.w3.org/TR/html5/embedded-content-0.html#media-controller-duration">media controller duration</a></code> to <var>new duration</var> and run the <a href="https://www.w3.org/TR/html5/embedded-content-0.html#durationChange">HTMLMediaElement duration change algorithm</a>.</li>
           </ol>
         </section>
@@ -1357,7 +1354,7 @@ interface <span class="idlInterfaceID">MediaSource</span> : <span class="idlSupe
                 <dd>
                   <ol>
                     <li>Run the <a href="#duration-change-algorithm">duration change algorithm</a> with <var>new duration</var> set to the highest end time reported by the <code><a href="#widl-SourceBuffer-buffered">buffered</a></code> attribute across all <a href="#idl-def-SourceBuffer" class="idlType"><code>SourceBuffer</code></a> objects in <code><a href="#widl-MediaSource-sourceBuffers">sourceBuffers</a></code>.<br>
-                      <div class="note"><div class="note-title marker" aria-level="5" role="heading" id="h-note16"><span>Note</span></div><p class="">This allows the duration to properly reflect the end of the appended media segments. For example, if the duration was explicitly set to 10 seconds and only media segments for 0 to 5 seconds were appended before endOfStream() was called, then the duration will get updated to 5 seconds.</p></div>
+                      <div class="note"><div class="note-title marker" aria-level="5" role="heading" id="h-note17"><span>Note</span></div><p class="">This allows the duration to properly reflect the end of the appended media segments. For example, if the duration was explicitly set to 10 seconds and only media segments for 0 to 5 seconds were appended before endOfStream() was called, then the duration will get updated to 5 seconds.</p></div>
                     </li>
                     <li>Notify the media element that it now has all of the media data.</li>
                   </ol>
@@ -1485,8 +1482,7 @@ interface <span class="idlInterfaceID">MediaSource</span> : <span class="idlSupe
           <p>On getting, Return the initial value or the last value that was successfully set.</p>
           <p>On setting, run the following steps:</p>
           <ol>
-            <li>If this object has been removed from the <code><a href="#widl-MediaSource-sourceBuffers">sourceBuffers</a></code> attribute of the <a href="#parent-media-source">parent media source</a>, then throw an <code><a href="https://www.w3.org/TR/html5/infrastructure.html#invalidstateerror">InvalidStateError</a></code> exception and
-              abort these steps.</li>
+            <li>If this object has been removed from the <code><a href="#widl-MediaSource-sourceBuffers">sourceBuffers</a></code> attribute of the <a href="#parent-media-source">parent media source</a>, then throw an <code><a href="https://www.w3.org/TR/html5/infrastructure.html#invalidstateerror">InvalidStateError</a></code> exception and abort these steps.</li>
             <li>If the <code><a href="#widl-SourceBuffer-updating">updating</a></code> attribute equals true, then throw an <code><a href="https://www.w3.org/TR/html5/infrastructure.html#invalidstateerror">InvalidStateError</a></code> exception and abort these steps.</li>
             <li>Let <var>new mode</var> equal the new value being assigned to this attribute.</li>
             <li>If <var><a href="#sourcebuffer-generate-timestamps-flag">generate timestamps flag</a></var> equals true and <var>new mode</var> equals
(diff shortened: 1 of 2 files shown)


jdsmith3000:
@wolenetz: Please review.  Addresses issues #19, #20 & #26.

paulbrucecotton:
@wolenetz;  Can you please review this PR ASAP so we can get issues #19, #20 & #26 dealt with?

wolenetz:
@paulbrucecotton @jdsmith3000 Yes - this is part of what I'm working through shortly.

jdsmith3000:
@wolenetz: Might you review by EoW?  Change implements our agreed upon solution. Addresses issues #19, #20 & #26.

wolenetz:
Yes, I'm looking at this today - got tied up earlier this week in trying out upstreaming a new web-platform-test. Reviews and PRs on spec issues will be coming faster now from me.

media-source-respec.html
```diff
@@ -829,19 +829,17 @@
           <p>Follow these steps when <a def-id="duration"></a> needs to change to a <var>new duration</var>.</p>
           <ol>
             <li>If the current value of <a def-id="duration"></a> is equal to <var>new duration</var>, then return.</li>
-            <li>Set <var>old duration</var> to the current value of <a def-id="duration"></a>.
+			<li>If <var>new duration</var> is less than the highest start time reported by the <a def-id="buffered"></a> attribute across all <a>SourceBuffer</a>
```
wolenetz: s/start time/end time/

media-source-respec.html
```diff
@@ -829,19 +829,17 @@
           <p>Follow these steps when <a def-id="duration"></a> needs to change to a <var>new duration</var>.</p>
           <ol>
             <li>If the current value of <a def-id="duration"></a> is equal to <var>new duration</var>, then return.</li>
-            <li>Set <var>old duration</var> to the current value of <a def-id="duration"></a>.
+			<li>If <var>new duration</var> is less than the highest start time reported by the <a def-id="buffered"></a> attribute across all <a>SourceBuffer</a> 
+			objects in <a def-id="sourceBuffers"></a>, then throw an <a def-id="invalid-state-error"></a> exception and abort these steps.</li>
+			  <p class="note">Duration reductions must first use <a def-id="remove"></a> to reduce the buffered range before updating <a def-id="duration"></a>.</p>
```
wolenetz: s/Duration reductions must use/Duration reductions that would truncate currently buffered media are disallowed. To accomplish such a truncation and duration reduction, use/

media-source-respec.html
```diff
@@ -829,19 +829,17 @@
...
+            <li>Set <var>old duration</var> to the current value of <a def-id="duration"></a>.</li>
             <li>Update <a def-id="duration"></a> to <var>new duration</var>.</li>
-            <li>If the <var>new duration</var> is less than <var>old duration</var>, then run the
-              <a def-id="range-removal"></a> algorithm with <var>new duration</var> and
-              <var>old duration</var> as the start and end of the removal range.
-
-              <p class="note">This preserves audio frames and text cues that start before and end after the <a def-id="duration"></a>.</p></li>
-            <li>If a user agent is unable to partially render audio frames or text cues that start before and end after the <a def-id="duration"></a>, then run the following steps:
+            <li>If a user agent is unable to partially render audio frames or text cues that start before and end after the <a def-id="duration"></a>, then run the following steps:</li>
+			  <p class="note">This condition can occur because the <a def-id="coded-frame-removal-algorithm"></a> preserves audio frames and text cues that start before and end after the <a def-id="duration"></a>.</p>
```
wolenetz: Hmm. It seems forcing `new duration` to be `>=` the highest end time across all SourceBuffer.buffered precludes this scenario. Am I incorrect?

media-source-respec.html
```diff
@@ -1078,6 +1076,7 @@
           <ol class="method-algorithm">
             <li>If this object has been removed from the <a def-id="sourceBuffers"></a> attribute of the <a def-id="parent-media-source"></a> then throw an <a def-id="invalid-state-error"></a> exception and abort these steps.</li>
             <li>If the <a def-id="readyState"></a> attribute of the <a def-id="parent-media-source"></a> is not in the <a def-id="open"></a> state then throw an <a def-id="invalid-state-error"></a> exception and abort these steps.</li>
+			<li>If the <a def-id="range-removal"></a> algorithm is running, then throw an <a def-id="invalid-state-error"></a> exception and abort these steps.</li>
```
wolenetz: nit: indentation in the respec.html isn't consistent with sibling `<li>`

media-source-respec.html
```diff
@@ -2723,7 +2722,17 @@
         </thead>
         <tbody>
           <tr>
-            <td>April 15 2016</td>
+            <td>April 18 2016</td>
```
wolenetz: nit/standard-operating-procedure reminder: Rebase, update date here, publishDate above, and regenerate index.html.

wolenetz:
Looks pretty good. See my comments, above. Let me know when they're addressed and I'll do a final review and merge this if it l*tm. Thanks!

media-source-respec.html
wolenetz: If I'm correct, this entire "partially render audio..." step, note, and substeps should be removed.

jdsmith3000: We've not changed the **Coded Frame Removal Algorithm** and the issue comes from it.

The Coded Frame Removal Algorithm takes starting and end removal range timestamps, and removes any data with a *starting timestamp* inside the removal range.

> Remove all media data, from this track buffer, that contain starting timestamps greater than or equal to start and less than the remove end timestamp.
 
Data that starts before the removal range and ends during it is not removed.  This means:

- Ending timestamps past duration are possible
- This affects all data types
- I'm not sure what the specific audio frame/text cue issue is, and decided to accommodate the mitigation for it.

I checked the start time in **Duration Change** on purpose because of this.  Then I left the audio/text cue special case to bump the duration if necessary.

wolenetz: highest start time across all SourceBuffer.buffered could be time 0 in a continuously buffered media timeline. That would certainly allow duration reduction to truncate complete buffered coded frames (which start after the highest buffered range start time). So I don't think that solves this problem.
I also see a problem now that SourceBuffer.buffered is defined as intersection ranges, so there could be (in certain scenarios where a muxed media segment for a portion of the timeline has not been completely appended prior to an abort(), for example) complete coded frames buffered for one track buffer in a SourceBuffer, which start at or after the highest end time of the (intersection ranges) buffered attribute for that SourceBuffer. Duration reduction would thereby require those frame's removal (asynchronously), and such asynchronous removal as part of duration reduction was what this PR and set of bugs was trying to fix.
How about instead of forcing new duration to be >= the highest (start or end) time across all SourceBuffer.buffered, we instead: force new duration to be >= the highest end time across all track buffer ranges across all track buffers across all SourceBuffers? This would meet the intent of this PR more cleanly, IMO. And since we've recently clarified track buffer ranges normatively in the spec to be "treated as if it were stored in a normalized TimeRanges object", no ambiguous definition remains IIUC. This would also allow us to remove the "bumping up" of new duration in the scenario steps this comment is anchored on in the PR. WDYT?

wolenetz: I do see also that "highest end time across all track buffers' track buffer ranges across all SourceBuffers" is a value not exposed to web apps. So still there's a problem even with my suggestion. The crux is SourceBuffer.buffered is defined in terms of intersection ranges, and muxed content could have complete coded frames in one track buffer that are outside the intersection ranges for that SourceBuffers. Any suggestions for how to resolve this?

wolenetz: On further thought (sorry for churn!) - using highest end time across buffered as the lower boundary for duration reductions means the "bumping up of duration" is redundant. It actually mimics the same behavior for updating duration that endOfStream() does. Why not just use highest end time across buffered as the boundary? (The muxed content scenario already exists in the case of jagged-ended appends: coded frame processing algorithm increases the duration if group end timestamp is higher, endOfStream() corrects this downwards to highest end time across buffered.)

jdsmith3000: You seem to be proposing changes to the error criteria for duration reductions that don't first do a `remove()`.  I believe to do that, we need to make a matching change to the **Coded Frame Removal Algorithm** so that `remove()` results will pass the check.

Separate from that, using latest start time as the check aligns with the current `remove()` algorithm.  Is there a problem with it currently that suggests broader change is needed?

wolenetz: > You seem to be proposing changes to the error criteria for duration reductions that don't first do a remove(). I believe to do that, we need to make a matching change to the Coded Frame Removal Algorithm so that remove() results will pass the check.

Changing the Coded Frame Removal Algorithm (CFRA) to comply with "highest buffered end time" check on subsequent duration reduction would probably regress some important playback scenarios: consider a long video frame or text cue that begin before the eventual new duration and end after it: we probably wouldn't want CFRA to remove those.

> Separate from that, using latest start time as the check aligns with the current remove() algorithm. Is there a problem with it currently that suggests broader change is needed?

I think this *might* be the route forward, however we need a clear definition in the spec of "the latest start time"; it's very rarely going to be the highest range start time in any SourceBuffer.buffered. And it's below the app-visible highest buffered end time. And it will still unfortunately require we "bump" the duration higher (following the case for platforms that can't handle partial frames/cues). I laid out a larger analysis that I can share, but it boiled down to: we need a precise definition for "highest start time", which I believe needs to be something like "highest buffered presentation timestamp across all SourceBuffers' track buffers" (HBPT, below). Ideally, this would be a MediaSource attribute exposed to a web app, so that it could make a reasonable decision in the following example cases where the app wants to change duration to X, where:
* X < HBPT: app should expect exception
* X >= HBPT and <= highest buffered end time across all SourceBuffers' track buffers' track ranges: app shouldn't expect exception, but actual resulting duration might instead be >X if the platform couldn't support partial audio frames/text cues.
* Otherwise, app should expect no exception or special casing as a result of spec changes here.

jdsmith3000: Your definition is reasonable, assuming presentation timestamp == starting timestamp in all cases.

I assume you want to expose HBPT to apps so they have a value that can be tested to determine if `remove()` should be run before a duration change is made, correct?  

I've assumed duration reductions in general will always require this (as they did when remove() was called automatically).  We are deleting the automatic `remove()` to solve some algorithm logic problems when the automatic `remove()` was made asynchronous.  Do you further believe apps will need to check whether remove() is needed on some reductions?

wolenetz: > assuming presentation timestamp == starting timestamp in all cases.

PTS is the presentation (start) time of a coded frame, in all cases as I understand). If, by starting timestamp, you instead meant range start time (and not coded frame PTS), that would be too broad a definition and require duration truncation to still need to remove buffered coded frames with PTS > "range start time".

> `[exposing HBPT to apps]`

I agree that remove() is a reasonable proactive or reactive (on duration truncation exception) response by an app when attempting to reduce duration, so HBPT need not be exposed to web apps. I was previously erroring on the side of not hiding critically needed info from apps, and as you point out, HBPT isn't such.

wolenetz:
@jdsmith3000 Yesterday, I responded to your most recent comments in the line notes, above. I believe we're in agreement on the route forward. Please let me know if you have any further questions. I'd like to review again once you've addressed the line notes, above. Thanks!

jdsmith3000:
@Wolenetz  Rebased commit picks up these changes based on your feedback:

-  Change Duration: used starting presentation timestamp in duration test
-  Expanded note on using remove()
-  Corrected indentation
-  Updated dates

jdsmith3000:
@Wolenetz Please review.  I think I've picked up the changes we settled on after discussion.  If so, let's finish this and close three issues.

wolenetz:
Yep - now that I'm back from my brief vacation, I'm reviewing both this and https://github.com/w3c/media-source/pull/73 today.

index.html
```diff
@@ -1318,19 +1318,17 @@ <h4 id="h-duration-change-algorithm" resource="#h-duration-change-algorithm"><sp
...
-            <li>If the <var>new duration</var> is less than <var>old duration</var>, then run the
-              <a href="#sourcebuffer-range-removal">range removal</a> algorithm with <var>new duration</var> and
-              <var>old duration</var> as the start and end of the removal range.
-
-              <div class="note"><div class="note-title marker" aria-level="5" role="heading" id="h-note15"><span>Note</span></div><p class="">This preserves audio frames and text cues that start before and end after the <code><a href="#widl-MediaSource-duration">duration</a></code>.</p></div></li>
-            <li>If a user agent is unable to partially render audio frames or text cues that start before and end after the <code><a href="#widl-MediaSource-duration">duration</a></code>, then run the following steps:
+			<li>If <var>new duration</var> is less than the highest starting <a href="#presentation-timestamp">presentation timestamp</a> reported by the <code><a href="#widl-SourceBuffer-buffered">buffered</a></code> attribute across all <a href="#idl-def-SourceBuffer" class="idlType"><code>SourceBuffer</code></a>
```
wolenetz: Please see https://github.com/w3c/media-source/pull/65#discussion_r61780279
Using "highest starting PTS reported by SB.buffered" as the allowed duration reduction lower bound is too lenient: it allows implicit removal of frames in that same buffered range with PTS > the range's "start" time. If that's not what you meant, this is really confusing.
I had thought we agreed that the "HBPT" would be the minimum allowable duration on attempted duration reduction, where HBPT==Highest buffered presentation time across all track buffers, across all SourceBuffers for this MediaSource. HBPT is not necessarily exposed to apps, and it is very rarely going to be the same as any timestamp reported by SourceBuffer.buffered.start(some range index), which is what this current text proposes to use (unless I misread).
Please change to:
If new duration is less than the highest presentation timestamp of any buffered coded frame across all track buffers and SourceBuffer objects for this MediaSource, then throw an ....

wolenetz: For a little more clarity on why this current phrasing could lead to implicit removal of the entire presentation interval of coded frame(s), consider the example single, un-muxed SourceBuffer whose .buffered reports one range beginning at time 0, but having 100 frames at 30ms intervals. If the app attempted to reduce the duration to 0, this phrasing would allow that without exception, but would implicitly remove all but perhaps the very first coded frame. This violates the intent of fixing these issues, which is to require explicit removal be done for any frame whose entire presentation interval lies after the new duration.

index.html
```diff
@@ -1318,19 +1318,17 @@ <h4 id="h-duration-change-algorithm" resource="#h-duration-change-algorithm"><sp
...
+			  <div class="note"><div class="note-title marker" aria-level="5" role="heading" id="h-note15"><span>Note</span></div><p class="">Duration reductions that would truncate currently buffered media are disallowed. When truncation is necessary, use <code><a href="#widl-SourceBuffer-remove-void-double-start-unrestricted-double-end">remove()</a></code> to reduce the buffered range before updating <code><a href="#widl-MediaSource-duration">duration</a></code>.</p></div>
+            <li>Set <var>old duration</var> to the current value of <code><a href="#widl-MediaSource-duration">duration</a></code>.</li>
+            <li>Update <code><a href="#widl-MediaSource-duration">duration</a></code> to <var>new duration</var>.</li>
+            <li>If a user agent is unable to partially render audio frames or text cues that start before and end after the <code><a href="#widl-MediaSource-duration">duration</a></code>, then run the following steps:</li>
```
wolenetz: replace MediaSource.duration with "new duration" (the variable).

media-source-respec.html
```diff
@@ -2731,11 +2729,13 @@
...
               <ul>
-                <li>Issue 54 - Clarified use of Presentation Time Stamps (PTS) for <a def-id="buffered"></a>.</li>
-              </ul>
+                <li>Issue 19, 20 &amp; 26 - Resolved problems with Duration Changes and Range Removals.</li>
+                <li>Separated Range Removals from Duration Changes, required separate remove().</li>
+                <li>Disabled abort on Range Removals.</li>
+             </ul>
             </td>
           </tr>
           <tr>
```
wolenetz: nit: please add the rawgit href to the previously-most-recent change.

jdsmith3000: It looks like Issue 54 was incorrectly removed from the Revision History.  I intend to restore it and add the rawgit link.

index.html
jdsmith3000: Is your intent to specify checking **all track buffers** in SourceBuffer objects for this MediaSource.  If so, I agree with that addition.  The language "for all SourceBuffer objects in SourceBuffers" seems equally correct to me though.

I agree that the intent of `remove()` is to remove all frames whose entire presentation interval lies after the new duration, and that the Duration Change condition should be aligned with this.

I'm puzzled by some of your issues though:

1. In your example, is your concern not removing the 0 time coded frame?  It should be removed, since it's timestamp is equal to the remove range start.
2. Are you concerned with using the `buffered` attribute to test the timestamps of buffered data?  I think it may not belong here, and that your language about timestamps for any buffered coded frame across all track buffers would be more correct.  

That might get us to:
> If new duration is less than the highest starting presentation timestamp of any buffered coded frame for all track buffers and SourceBuffer objects in sourceBuffers, then throw an InvalidStateError exception and abort these steps.

This leaves in "starting" to align with this Coded Frame Removal language:

> Remove all media data, from this track buffer, that contain starting timestamps greater than or equal to start and less than the remove end timestamp.

So if we settle whether this is correct or not, I think we can proceed.

wolenetz: > Is your intent to specify checking all track buffers in SourceBuffer objects for this MediaSource. If so, I agree with that addition. The language "for all SourceBuffer objects in SourceBuffers" seems equally correct to me though.

Yes. "all track buffers in SourceBuffer objects in SourceBuffers" is slightly more precise than "all SourceBuffer objects in SourceBuffers". Both are correct.

> I agree that the intent of remove() is to remove all frames whose entire presentation interval lies after the new duration, and that the Duration Change condition should be aligned with this.

Agreed.

> 1. In your example, is your concern not removing the 0 time coded frame? It should be removed, since it's timestamp is equal to the remove range start.

I didn't mean to confuse. My concern was that any later coded frames (those with higher presentation timestamps than "0" in the example) would need to be implicitly removed to align with the Duration Change condition, and thereby such a duration reduction should be disallowed.

> 2. Are you concerned with using the buffered attribute to test the timestamps of buffered data?

Yes, That is my concern.

> I think it may not belong here, and that your language about timestamps for any buffered coded frame across all track buffers would be more correct.

Yes.

Your proposed new wording (*) looks good to me. I think we're now on the same page again. Let's proceed :)

(*) If new duration is less than the highest starting presentation timestamp of any buffered coded frame for all track buffers and SourceBuffer objects in sourceBuffers, then throw an InvalidStateError exception and abort these steps.

jdsmith3000:
@wolenetz I believe this is now ready to merge.

wolenetz:
Looks good! Since the EME-discussion about the GH "green button" hasn't settled / I'm not sure I trust the button yet, I'll go ahead and merge this manually, shortly. Thanks for getting this updated, @jdsmith3000 

Note to web authors and UA implementors: This spec change may break MSE players which previously depended upon implicit (and ambiguously defined) range removals on duration reductions which would truncate buffered media, and similarly might break players which attempt to abort() asynchronous range-removals. The workaround for the former is now a well-defined explicit remove(), and for the latter, there is no longer allowed any ability for an app to abort() range removals.

plehegar:
I'm curious: does this change break existing UA implementations and, if yes, do we know the timeline for them to get aligned?

jdsmith3000:
Yes, this will require implementations to update.  The previous spec was not actually implementable however.

We need to discuss implementation updates in general, since this is not the only issue that could require updates.

PR #72, opened 2016-04-29
Fixed broken fragments

plehegar:
Meant to send those fix last time and it didn't, so here they are.


plehegar:
doh. duplicates #38...

PR #73, opened 2016-05-02
Fix #15 and some minor typos

### LINKED ISSUE #15: Clarify how track buffer ranges are updated.
https://www.w3.org/Bugs/Public/show_bug.cgi?id=27242

wolenetz:
- Clarified track buffer ranges.
- Clarified allowance near presentation start time.
- Minor whitespace and formatting fixes.

### DIFF
--- a/index.html
+++ b/index.html
@@ -442,7 +442,7 @@ table.simple {
     }
 }
 </style><link rel="stylesheet" href="https://www.w3.org/StyleSheets/TR/2016/W3C-ED"><!--[if lt IE 9]><script src='https://www.w3.org/2008/site/js/html5shiv.js'></script><![endif]--></head>
-  <body class="h-entry" role="document" id="respecDocument"><div class="head" role="contentinfo" id="respecHeader">
+  <body class="h-entry toc-inline" role="document" id="respecDocument"><div class="head" role="contentinfo" id="respecHeader">
   <p>
       
         
@@ -812,6 +812,8 @@ table.simple {
           <dt id="presentation-start-time">Presentation Start Time</dt>
           <dd><p>The presentation start time is the earliest time point in the presentation and specifies the <a href="https://www.w3.org/TR/html5/embedded-content-0.html#initial-playback-position">initial playback position</a> and <a href="https://www.w3.org/TR/html5/embedded-content-0.html#earliest-possible-position">earliest possible position</a>. All presentations created using this specification have a presentation start time of 0.</p></dd>
 
+          <div class="note"><div class="note-title marker" aria-level="4" role="heading" id="h-note4"><span>Note</span></div><p class="">For the purposes of determining if <code><a href="https://www.w3.org/TR/html5/embedded-content-0.html#dom-media-buffered">HTMLMediaElement.buffered</a></code> contains a <code><a href="https://www.w3.org/TR/html5/embedded-content-0.html#timeranges">TimeRange</a></code> that includes the current playback position, implementations may choose to allow a current playback position at or after <a href="#presentation-start-time">presentation start time</a> and before the first <code><a href="https://www.w3.org/TR/html5/embedded-content-0.html#timeranges">TimeRange</a></code> to play the first <code><a href="https://www.w3.org/TR/html5/embedded-content-0.html#timeranges">TimeRange</a></code> if that <code><a href="https://www.w3.org/TR/html5/embedded-content-0.html#timeranges">TimeRange</a></code> starts within a reasonably short time, like 1 second, after <a href="#presentation-start-time">presentation start time</a>. This allowance accommodates the reality that muxed streams commonly do not begin all tracks precisely at <a href="#presentation-start-time">presentation start time</a>. Implementations should report the actual buffered range, regardless of this allowance.</p></div>
+
           <dt id="presentation-interval">Presentation Interval</dt>
           <dd>
             <p>The presentation interval of a <a href="#coded-frame">coded frame</a> is the time interval from its <a href="#presentation-timestamp">presentation timestamp</a> to the <a href="#presentation-timestamp">presentation timestamp</a> plus the <a href="#coded-frame-duration">coded frame's duration</a>. For example, if a coded frame has a presentation timestamp of 10 seconds and a <a href="#coded-frame-duration">coded frame duration</a> of 100 milliseconds, then the presentation interval would be [10-10.1). Note that the start of the range is inclusive, but the end of the range is exclusive.</p>
@@ -880,10 +882,10 @@ table.simple {
     "<a href="#idl-def-EndOfStreamError.decode" class="idlEnumItem">decode</a>"
 };</span></pre><table class="simple"><tbody><tr><th colspan="2">Enumeration description</th></tr><tr><td><code id="idl-def-EndOfStreamError.network">network</code></td><td>
           <p>Terminates playback and signals that a network error has occured.</p>
-          <div class="note"><div class="note-title marker" aria-level="3" role="heading" id="h-note4"><span>Note</span></div><p class="">JavaScript applications should use this status code to terminate playback with a network error. For example, if a network error occurs while fetching media data.</p></div>
+          <div class="note"><div class="note-title marker" aria-level="3" role="heading" id="h-note5"><span>Note</span></div><p class="">JavaScript applications should use this status code to terminate playback with a network error. For example, if a network error occurs while fetching media data.</p></div>
         </td></tr><tr><td><code id="idl-def-EndOfStreamError.decode">decode</code></td><td>
           <p>Terminates playback and signals that a decoding error has occured.</p>
-          <div class="note"><div class="note-title marker" aria-level="3" role="heading" id="h-note5"><span>Note</span></div><p class="">JavaScript applications should use this status code to terminate playback with a decode error. For example, if a parsing error occurs while processing out-of-band media data.</p></div>
+          <div class="note"><div class="note-title marker" aria-level="3" role="heading" id="h-note6"><span>Note</span></div><p class="">JavaScript applications should use this status code to terminate playback with a decode error. For example, if a parsing error occurs while processing out-of-band media data.</p></div>
         </td></tr></tbody></table>
 
       <pre class="idl"><span class="idlInterface" id="idl-def-MediaSource">[<span class="extAttr">Constructor</span>]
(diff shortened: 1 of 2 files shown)


wolenetz:

index.html
```diff
@@ -1613,11 +1617,13 @@ <h3 id="h-track-buffers" resource="#h-track-buffers"><span property="xhv:role" r
...
+          The threshold for determining disjointness of <a href="#track-buffer-ranges">track buffer ranges</a> is implementation-specific.
+          To approximate the discontinuity detection logic within the <a href="#sourcebuffer-coded-frame-processing">coded frame processing algorithm</a>, it is recommended when determining if one time range is continuous with another as a result of newly processed or removed coded frames, to allow continuity across no more than twice the maximum frame duration seen so far by this <a href="#track-buffer">track buffer</a> between the earlier range's end timestamp and the later range's earliest <a href="#presentation-timestamp">presentation timestamp</a>. In a muxed <a href="#idl-def-SourceBuffer" class="idlType"><code>SourceBuffer</code></a>, implementations may choose to use the earliest <a href="#presentation-timestamp">presentation timestamp</a> across all <a href="#track-buffer">track buffers</a> following a discontinuity detected by <a href="#sourcebuffer-coded-frame-processing">coded frame processing algorithm</a> as the range start for the track buffers across that muxed group of frames.</p></div>
```
wolenetz: Note, Chrome is just a little more strict than this. If max frame duration seen so far is "10", Chrome would find the following to be discontinuous (notation: {coded frame PTS, duration}, assuming everything is a keyframe for simplicity in these examples):
{0,10} and {21,10} would be discontinuous ranges [0,10) [21,31)
{0,10} and {19,10} would be continuous [0,29)
This is because Chrome operates on the delta (more like the coded frame processing algorithm) between the earlier range's highest PTS (not end timestamp) and the next range's earliest PTS being within 2 * max duration seen so far.  (Note that Chrome uses DTS incorrectly still, but ignore that for the purposes of this discussion.) There are still some hairy edge cases that don't conform to even this regime in Chrome (like the frame duration used in calculating the max seen so far isn't precisely that in all cases), hence the "implementation-specific" and non-normative note framing of this.

jdsmith3000:
The additions under track buffer ranges seem like they could be clearer.  We believe the comment on presentation time ranges is best handled in a non-normative note rather than in the normative paragraph.

> These presentation time ranges are based on presentation timestamps, frame durations, and potentially coded frame group start times for coded frame groups across track buffers in a muxed SourceBuffer.

We also believe the note expanding on the coded frame group start times could be simplified.  It's pretty complex, but seems to be saying that implementations should ignore a gap in the buffered ranges provided the gap is not larger than 2 times the max frame duration seen so far.  Is that correct?  Can it be simplified then to something closer to that?

> These coded frame group start times differ slightly from those mentioned in the coded frame processing algorithm in that they are the earliest presentation timestamp across all track buffers following a discontinuity, which can occur within the coded frame processing algorithm or result from the coded frame removal algorithm, regardless of mode. The threshold for determining disjointness of track buffer ranges is implementation-specific. To approximate the discontinuity detection logic within the coded frame processing algorithm, it is recommended when determining if one time range is continuous with another as a result of newly processed or removed coded frames, to allow continuity across no more than twice the maximum frame duration seen so far by this track buffer between the earlier range's end timestamp and the later range's earliest presentation timestamp. In a muxed SourceBuffer, implementations may choose to use the earliest presentation timestamp across all track buffers following a discontinuity detected by coded frame processing algorithm as the range start for the track buffers across that muxed group of frames.

wolenetz:
There was a stray extra copy/paste snippet earlier in the respec.html which I'll remove.

Regarding:

> The additions under track buffer ranges seem like they could be clearer. We believe the comment on presentation time ranges is best handled in a non-normative note rather than in the normative paragraph.

> > These presentation time ranges are based on presentation timestamps, frame durations, and potentially coded frame group start times for coded frame groups across track buffers in a muxed SourceBuffer.

I'll move this portion back to non-normative note, since "presentation time ranges" will remain in the normative text.

wolenetz:
Regarding:

> We also believe the note expanding on the coded frame group start times could be simplified. It's pretty complex, but seems to be saying that implementations should ignore a gap in the buffered ranges provided the gap is not larger than 2 times the max frame duration seen so far. Is that correct? Can it be simplified then to something closer to that?

"coded frame group start times" is very similar to "group start time" CFP variable, which is valid only ephemerally in 'sequence' AppendMode CFP processing. Differentiating the two was why the first part of this new note was a bit verbose. I think making the "To approximate..." portion of the new note simpler would help though -- I'll update this PR accordingly shortly.

wolenetz:
@jdsmith3000 please take a look. I've tried to make the new notes more clear and concise per your previous comments. If you agree that it's ready to land, let me know and I'll merge it. Thanks!

media-source-respec.html
```diff
@@ -309,6 +309,8 @@
...
           <dd><p>The presentation start time is the earliest time point in the presentation and specifies the <a def-id="videoref" name="initial-playback-position">initial playback position</a> and <a def-id="videoref" name="earliest-possible-position">earliest possible position</a>. All presentations created using this specification have a presentation start time of 0.</dd>
 
+          <p class="note">For the purposes of determining if <a def-id="hme-buffered"></a> contains a <a def-id="timerange"></a> that includes the current playback position, implementations may choose to allow a current playback position at or after <a def-id="presentation-start-time"></a> and before the first <a def-id="timerange"></a> to play the first <a def-id="timerange"></a> if that <a def-id="timerange"></a> starts within a reasonably short time, like 1 second, after <a def-id="presentation-start-time"></a>. This allowance accommodates the reality that muxed streams commonly do not begin all tracks precisely at <a def-id="presentation-start-time"></a>. Implementations should report the actual buffered range, regardless of this allowance.</p>
+
```
jdsmith3000: This note addresses the jagged start concern, and looks good.

media-source-respec.html
```diff
@@ -1199,11 +1201,14 @@
...
+        <p>Each <a def-id="track-buffer"></a> has a <dfn id="track-buffer-ranges">track buffer ranges</dfn> variable that represents the presentation time ranges occupied by the <a def-id="coded-frames"></a> currently stored in the track buffer.
+
+          <p class="note">For track buffer ranges, these presentation time ranges are based on <a def-id="presentation-timestamp">presentation timestamps</a>, frame durations, and potentially coded frame group start times for coded frame groups across track buffers in a muxed <a>SourceBuffer</a>.</p>
 
-        <p class="note">For track buffer ranges, these "presentation time ranges" are based on Presentation Time Stamps (PTS), frame durations, and potentially coded frame group start times for coded frame groups across track buffers in a muxed <a>SourceBuffer</a>.</p>
+          For specification purposes, this information is treated as if it were stored in a <a def-id="normalized-timeranges-object"></a>. Intersected <a def-id="track-buffer-ranges"></a> are used to report <a def-id="hme-buffered"></a>, and must therefore support uninterrupted playback within each range of <a def-id="hme-buffered"></a>.
```
jdsmith3000: Restored to non-normative note that we'd previously discussed.  Looks good..

media-source-respec.html
```diff
@@ -1199,11 +1201,14 @@
...
 
+          <p class="note">These coded frame group start times differ slightly from those mentioned in the <a def-id="coded-frame-processing-algorithm"></a> in that they are the earliest <a def-id="presentation-timestamp"></a> across all track buffers following a discontinuity, which can occur within the <a def-id="coded-frame-processing-algorithm"></a> or result from the <a def-id="coded-frame-removal-algorithm"></a>, regardless of <a def-id="mode"></a>.
```
jdsmith3000: Might the first sentence break after "discontinuity"?

> ...across all tracks following a discontinuity.  Discontinuities can occur....

Or am I misreading your intent?

wolenetz: Yes, I can split the sentence there with a period as you suggest and still retain (and improve clarity/communication of) the intended meaning.

media-source-respec.html
```diff
@@ -1199,11 +1201,14 @@
...
+          The threshold for determining disjointness of <a def-id="track-buffer-ranges"></a> is implementation-specific; to reduce unexpected playback stalls, implementations may approximate the <a def-id="coded-frame-processing-algorithm"></a>'s discontinuity detection logic by coalescing adjacent ranges separated by a gap smaller than 2 times the maximum frame duration buffered so far in this <a def-id="track-buffer"></a>. Implementations may also use coded frame group start times as range start times across <a def-id="track-buffer">track buffers</a> in a muxed <a>SourceBuffer</a> to further reduce unexpected playback stalls.</p>
```
jdsmith3000: Might the first sentence here break at the ";"?

> ...disjointness of track buffer ranges is implementation-specific.  To reduce unexpected playback stalls...

Is the approximation of the discontinuity detection logic actually derived somehow from the coded frame processing algorithm?  I'm not familiar with it being:

> 2 times the maximum frame duration buffered so far in this track buffer.

wolenetz:
I'll rebase on top of b386c1c (current gh-pages) and split that sentence as described at https://github.com/w3c/media-source/pull/73#discussion_r63100167

media-source-respec.html
wolenetz: > [ break at the ";" ? ]
Sounds good to me. I'll do that shortly.

> [ Approximation's derivation? ] 
There is no "maximum frame duration buffered so far" in the coded frame processing algorithm (CFP). However, CFP detects discontinuities when the DTS sequence goes backwards (inapplicable to this approximation derivation) or when:
"If last decode timestamp for track buffer is set and the difference between decode timestamp and last decode timestamp is greater than 2 times last frame duration:"
Note especially the "2 times last frame duration" portion of this. Also note that this approximation text is fully within a non-normative note, and is intended to help implementations and web authors understand that there is lenience (the threshold...is implementation-specific) and that there are some suggestions for reaching better interop here. Would you prefer different, or even normative, text, for this?

wolenetz:
Actually, I'm pending @jdsmith3000 first responding to https://github.com/w3c/media-source/pull/73#discussion_r63101828 before I'll update the associated commit for this PR.

media-source-respec.html
wolenetz: > > [ break at the ";" ? ]
> Sounds good to me. I'll do that shortly.
After the sentence break, I'll add "For example, " to reduce unexpected playback stalls.

wolenetz:
fb01530 is current, and the only pending item is @jdsmith3000 's response to https://github.com/w3c/media-source/pull/73#discussion_r63101828. @jdsmith3000, does this look good?

jdsmith3000:
@wolenetz I'm okay with this language, and prefer to keep it non-normative.  There's probably room to refine our behavior handling gaps, but I don't think that's a V1 topic.  Your explanation of the "2 times last frame duration" logic makes sense to me as well.

I think this is ready to merge.

wolenetz:
Thanks for the review. I'll merge this shortly.

PR #81, opened 2016-05-26
Fix #68

### LINKED ISSUE #68: Array bracket syntax is deprecated
The attribute definition

```
readonly        attribute DOMString[]      kinds;
```
in the interface `TrackDefault` should be changed to something like

```
readonly        attribute sequence<DOMString>      kinds;
```

plehegar:
kinds attribute should be a sequence (and match the constructor definition)


wolenetz:
Thanks for preparing this! I landed this yesterday as c259f4f1a79ad7306ed258090234d99eea14f9e4 (prior to seeing this PR generated) :)

PR #82, opened 2016-05-26
Fix #44

### LINKED ISSUE #44: Inaccessible diagram
From
  https://lists.w3.org/Archives/Public/public-html-media/2016Jan/0000.html

[[
The diagram in this document should be made accessible. There
are several ways to achieve this, including accessible vector graphics
with SVG, a link to an external description (standard <a href> or
@longdesc), or an in-page description of the graphical content.

	a.)	We provide for your reference a draft textual diagram
	description, though our draft requires enhancement as per b.)
	below. The draft description is available at:

	http://lists.w3.org/Archives/Public/public-apa/2016Jan/0014.html

	b.)	In addition to a textual description of your diagram for
	the benefit of those readers who cannot see it, wWe find the
	diagram also does not clearly convey its meaning and may benefit
	significantly from redesign. Please reconsider what you're
	trying to convey and determine if there is a more intuitive way
	to present the data.  Please clearly label your switches,
	connector endpoints, and relationships, e.g. What are the three
	dots under the video decoders? Are all the audio decoder
	switches open or is that a drawing error? What is the green
	circled X that the audio decoders are pointing to? What are the
	triangles? What is the meaning of the different colors of
	SourceBuffers? What is the viewer expected to learn from this
	diagram?

	c.)	Please consider creating this diagram using SVG. We
	believe it is actually an excellent candidate for the accessible
	graphic markup approaches being defined by the joint ARIA and
	SVG Task Force.[2] One member of this TF has already offered to
	assist, should you decide on SVG.
]]

plehegar:
Pipeline figure longdesc svg


plehegar:
https://github.com/w3c/media-source/issues/44

wolenetz:
Thanks for preparing this! I landed this yesterday as 1f80b6e98d13acc106d5da80a771b6e4756d971c (prior to seeing this PR generated) :)

PR #86, opened 2016-05-28
Fix #18: Add fetch failures to MediaSource detachment note

### LINKED ISSUE #18: Should fetch algorithm failure trigger detaching from a media element?

wolenetz:
@jdsmith3000 I believe this implements the spec change to fix #18 per our discussion yesterday. Please take a look.

@foolip Please also take a look. I'll be making a Chromium/Blink CL shortly to see what this looks like in our implementation.

Note to reviewers: Please don't use the "big green merge" button. Instead, let me do the necessary rebasing and land this once it looks good to you.

Thanks!

### DIFF
--- a/index.html
+++ b/index.html
@@ -1162,7 +1162,7 @@ interface <span class="idlInterfaceID">MediaSource</span> : <span class="idlSupe
             <li>
               <a href="https://www.w3.org/TR/html5/webappapis.html#queue-a-task">Queue a task</a> to <a href="https://www.w3.org/TR/html5/webappapis.html#fire-a-simple-event">fire a simple event</a> named <code><a href="#dom-evt-sourceclose">sourceclose</a></code> at the <a href="#idl-def-MediaSource" class="idlType"><code>MediaSource</code></a>.</li>
           </ol>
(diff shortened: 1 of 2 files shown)


foolip:
A bit tricky to review with one long line, but it looks like all of this is in a non-normative note and thus isn't what implementors should actually use to decide what to do. Once the intended behavior is a bit more clear in https://github.com/w3c/media-source/issues/18 I hope we can get the hooks into one or both HTML specs.

jdsmith3000:
LGTM.

This change implements the note wording we defined in our discussion.  There were numerous fetch failure cases in HTML5.1 and listing them did not seem appropriate.  This change adds them as a group to the previously existing note and exempts the fetch failure that doesn't require detachment.

> Resource fetch algorithm failures are those which abort either the resource fetch algorithm or the resource selection algorithm, with the exception that the "Final step" [HTML51] is not considered a failure that triggers detachment.

PR #92, opened 2016-06-09
Fix #5: Add support for customizing seekable on live MSE streams

### LINKED ISSUE #5: Seekable differs from non-MSE behavior

wolenetz:


wolenetz:
@jdsmith3000 @mwatson2 @paulbrucecotton @plehegar FYI - this PR is ready for review.

@jdsmith3000 Note that even with the "live seekable range" set, the new related seekable logic (used only if duration is +Infinity) will return a TimeRanges object with precisely 1 range that spans the earliest buffered or live seekable range start time to the highest buffered to seekable range end time. I suppose we could make this a little more obviously not require a full "union" computation, but implementors are welcome to optimize to meet the requirements. WDYT? I also felt that either an empty seekable or a single-range-in-seekable for MSE case was the simplest change that solved two previous problems with +Infinity duration MSE in #5:
1) Seekable no longer needs to begin at time 0. An app can override that. Note that endOfStream() makes duration finite, so the old path would then become engaged and seekable would be [0,duration) (or an empty range). For live-streaming MSE players, I don't think endOfStream() is likely to be desired. WDYT?
2) Seekable now can include positions beyond the highest buffered end time, allowing live streaming apps to seek beyond buffered end time (perhaps all the way to the current live edge or beyond). Again, endOfStream() negates this customization.

One other note: An app could conceivably attach MediaSource to HTMLMediaElement and then explicitly set its duration to +Infinity prior to reaching HAVE_METADATA. The app could then setLiveSeekableRange() to allow MSE seeking to not potentially lose a nonzero seek target time that might be requested prior to reaching HAVE_METADATA. This is a further app issue that I believe this spec change will make it more easy to solve.

jdsmith3000:
LGTM

This seems to do exactly what we agreed upon.  The text you added to MediaSource Object outlines the purpose of live seekable range and defines set and clear methods that work on it.  These seem correctly described as well, and the resulting adjustments to HTMLMediaElement Extensions apply the value only to live streams, union the live seekable range with buffered when it's not empty, and use just buffered if it is.  It's good by me.

wolenetz:
Thanks @jdsmith3000 - I'll merge this now.

PR #114, opened 2016-07-12
Fix #52: Adds conformance section and uses RFC2119 key words

### LINKED ISSUE #52: Use RFC 2119 words
ReSpec supports [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt) words ("MUST", "MUST NOT", etc.). If such words are capitalized, ReSpec will format them in a way that indicates their importance. Neither MSE or the Byte Stream Format pages appear to use these.

wolenetz:


wolenetz:

Note, not all of these changes are simple capitalization of previous key words. I changed a few, like MAY->SHOULD in at least one place. Also, my other currently pending PRs may include some conflicts which will need resolution when I get to landing this.

jdsmith3000:
LGTM @Wolenetz.  I noted one MUST that became a SHOULD as well, but agree with the change.

wolenetz:
Thanks for the review, @jdsmith3000. I'll merge this shortly (and adjust those other PRs that have merge conflicts when merging them once you've approved them).

For posterity, these are the specific changes (other than rote capitalization and conformance section/biblio inclusion, and one replacement of "&" with "and"):

1) 
...The presentation timestamp in a <a def-id="coded-frame"></a> indicates when the frame must be rendered....
---> s/must/SHOULD/ since there could be decoder stalls or other reasons why the user agent might drop the frame instead of rendering it during playback.

2)
...Implementations should report the actual buffered range, regardless of this allowance....
---> s/should/MUST/ since yes, implementations really MUST here :)

3)
...An <a def-id="addSourceBuffer"></a> call may still fail if sufficient resources are not available to support the addition of a new <a>SourceBuffer</a>....
---> s/may/SHOULD/ since SHOULD makes more sense to me here...

4)
...<p>The following steps are run in any case where the media element is going to transition to <a def-id="videoref" name="dom-htmlmediaelement-network_empty">NETWORK_EMPTY</a> and <a def-id="queue-a-task-to-fire-an-event-named"></a> <a def-id="videoref" name="eventdef-media-emptied">emptied</a> at the media element. These steps must be run right before the transition.</p>...
---> s/must/SHOULD/ since we have that non-normative note that allows variance as HTML51 and beyond take refinements around detachment scenarios.

PR #119, opened 2016-07-13
Fix #118: Remove restriction on not 'updating' from {set,clear}LiveSeekableRange()

### LINKED ISSUE #118: Restriction on not 'updating' in {set,clear}LiveSeekableRange is too strict
At least one MSE web author has reported that the step "If the updating attribute equals true on any SourceBuffer in sourceBuffers, then throw an InvalidStateError exception and abort these steps" in {set,clear}LiveSeekableRange makes the feature very hard to use and doesn't add value.

Consider the case where a live streaming app is appending small media segments independently to multiple SourceBuffers for one MediaSource, and the app wishes to tightly maintain the live seekable range. It's difficult for such an app to coordinate the calls to set/clearLiveSeekableRange to occur when 'updating' isn't in progress. Furthermore, the underlying usage of the "live seekable range" in the extended HTMLMediaElement.seekable logic itself makes no restriction on 'updating' not being in progress (nor should it).

This issue tracks removing the 'updating' check from {set,clear}LiveSeekableRange().

@jdsmith3000 I hope this doesn't constitute a substantive change -- it would be excellent if this issue were fixed in the current V1 CR time-frame. I'll prepare a candidate pull request for this today.

I'm triaging this to V1Editorial. Please speak up here if you disagree and believe this is substantive.

wolenetz:


wolenetz:
@jdsmith3000 Please take a look. I'd like to land this ASAP since it's borderline substantive. This is intended to fix #118 and make the {set,clear}LiveSeekableRange() API more broadly usable by removing the unnecessary restriction on not 'updating' from those methods.

jdsmith3000:
In hindsight, I'm not sure why we put in this restriction initially.  It's a fair criticism that the values might be frequently updated, and the underlying usage doesn't really impose a need for avoiding updates.

The change is clean.  LGTM.

wolenetz:
Thanks for the review. I'll manually merge this shortly.

PR #122, opened 2016-07-14
Fix #49: Update byte stream format registry and spec headers, sotd, bug links, etc.

wolenetz:
Includes previously missing sotd and abstract sections where necessary,
and removes unnecessary empty "SourceBuffer" entries in the local
respec definitionMap to net 0 respec errors for the updated respec
files in this change.

PR #117, opened 2016-07-12
Fix #115: Add conformance section and use RFC2119 in byte stream formats and registry

wolenetz:
The main MSE spec will later need a biblio update to reflect
the editor updates included in the byte stream format registry in this
change. The registry also includes a WG name update.

The isobmff, mp2t, and mpeg audio byte stream format specs include
editor and WG name updates (the webm format spec already had those
updates.)

PR #125, opened 2016-07-19
Fix #50: Move avc1-4 to non-normative note

### LINKED ISSUE #50: ISO BMFF Byte Stream Format requires avc1-4 support
https://w3c.github.io/media-source/isobmff-byte-stream-format.html#iso-init-segments normatively says:
>The user agent must support parameter sets (e.g., PPS/SPS) stored in the sample entry (as defined for avc1/avc2), and should support parameter sets stored inband in the samples themselves (as defined for avc3/avc4).

There are several issues:
* User agents are required to support parameter sets.
* User agents must do so according to avc1-4.
* An example ("e.g.") is provided in normative text.

Ideally, the format specification would not need to be updated to handle new specifications. Is there a way to accomplish the purpose of this text in another way? Perhaps an example in a non-normative Note will provide the necessary context for a more general statement.

wolenetz:


wolenetz:
@jdsmith3000 Please review this. @ddorwin, I believe this should fix #50 - please also review this since you filed that issue originally. Thanks, both!

p.s. There are other pending PRs impacting these files still pending review: #122 and #117. I'll resolve conflicts when I land these later.

isobmff-byte-stream-format-respec.html
```diff
@@ -148,9 +148,9 @@
         (<span class="iso-box">elst</span>) that contains a single edit with media rate one. This edit may have a duration of 0 (indicating that it spans all subsequent media) or may have a non-zero duration
         (indicating the total duration of the movie including fragments).</p>
 
-      <p>The user agent must support parameter sets (e.g., PPS/SPS) stored in the sample entry (as defined for avc1/avc2), and should support parameter
-        sets stored inband in the samples themselves (as defined for avc3/avc4).</p>
-      <p class="note">For maximum content interoperability, user agents are strongly advised to support both inband and out-of-band storage of the SPS and PPS.</p>
+      <p>The user agent must support parameter sets (i.e., PPS/SPS) stored out-of-band in the sample entry, and should support parameter
```
ddorwin: Should PPS/SPS be moved to the note too? Are those related to avc1/2? Perhaps restate some of this in the form of "For example, PPS/SPS in avc1/avc2..." (this is very rough).

ddorwin: Assuming PPS/SPS are avc-specific, "i.e." does not seem appropriate either.

wolenetz: You're right. PPS/SPS are specific to h264, not ISOBMFF in general. I'll try another iteration here to make this even more generic.

wolenetz: Take a look at fa22c43, which tries to make this more generic.

jdsmith3000:
I like this better with you recent change.  LGTM.

wolenetz:
Thanks for the review. I'll manually merge this shortly.

wolenetz:
It looks like my previous landing of  #117 didn't include a regenerated (with updated date) isobmff registry. This will be fixed when I merge this PR.

Issue #126, opened 2016-07-19
Consider removing the revision histories


wolenetz:
See https://github.com/w3c/encrypted-media/issues/280

Pending guidance for EME (and MSE) from @plehagar for this issue originally reported by @ddorwin for EME.

It might be more useful to retain revision history in just the MSE bytestream specs and registry, if we do decide to remove that history from the main MSE spec doc.

jdsmith3000:
I support removing this on both EME (where it hasn't been updated) and MSE (where it has).
As an alternative, the Web Audio Spec just pulls in a GitHub change log: http://www.w3.org/TR/webaudio/changelog.html

wolenetz:
MSE and byte stream format registry and specs all now include a "commit history" link in the document headers.

I'll prep a PR to remove the revision history section from the 2 MSE specs while contain in: (the main spec, and the ISOBMFF bytestream format spec).

PR #128, opened 2016-07-21
Fix #93: Origin of MediaSource object URL should be relevant settings object of this when it is created

### LINKED ISSUE #93: "effective script origin" is no longer a thing
And it seems very strange to use that anyway, as it only matters for some legacy `document.domain` cases.

mwatson2:


jdsmith3000:
Needs an entry in Revision History (which is maintained for each fixed issue in MSE).  The change itself looks fine.

We link in rawgit versions for the previous change as well, but should be able to catch up on that on a subsequent change.

index.html
```diff
@@ -802,7 +802,7 @@ <h3 id="h-definitions" resource="#h-definitions"><span property="xhv:role" resou
...
             <p>These URLs are the same as a <a href="https://www.w3.org/TR/FileAPI/#url">Blob URI</a>, except that anything in the definition of that feature that refers to <a href="https://www.w3.org/TR/FileAPI/#dfn-file">File</a> and <a href="https://www.w3.org/TR/FileAPI/#dfn-Blob">Blob</a> objects is hereby extended to also apply to <a href="#idl-def-MediaSource" class="idlType"><code>MediaSource</code></a> objects.</p>
-            <p>The <a href="https://www.w3.org/TR/html51/browsers.html#origin">origin</a> of the MediaSource object URL is the <a href="https://www.w3.org/TR/html51/browsers.html#effective-script-origin">effective script origin</a> of the document that called <code><a href="#widl-URL-createObjectURL-DOMString-MediaSource-mediaSource">createObjectURL()</a></code>.</p>
+            <p>The <a href="https://www.w3.org/TR/html51/browsers.html#origin">origin</a> of the MediaSource object URL is the <a href="https://www.w3.org/TR/html51/webappapis.html#relevant-settings-object">relevant settings object</a> of <code>this</code> during the call to <code><a href="#widl-URL-createObjectURL-DOMString-MediaSource-mediaSource">createObjectURL()</a></code>.</p>
```
wolenetz: So I understand more clearly, what is "this" during the call to URL.createObjectURL(MediaSource), and what is the relevant settings object for that `this`?
Does the MediaSource argument have a relevant settings object which has an origin that is appropriate here, or are we instead saying that the URL interface object's relevant settings object should be used (is that a distinction without a difference?)

wolenetz:
Regarding revision history, if #126 lands first, then such revision history update will be unnecessary.
publishDate **will** still need to be updated just prior to merging though.

index.html
mwatson2: Well, my understanding of this issue is admittedly not complete, but my understanding is that, "this" here is the URL interface object and the "relevant settings object" has a 1-1 relationship with the current global scope / window object. The MediaSource object would indeed have a relevant settings object which would be associated with the global scope in which it was created.

Certainly, a MediaSource object with a different relevant settings object could exist, but what I am not sure about is whether you could get such a thing into a place where you could call URL.createObjectURL on it with a URL object connected to a different global.

mwatson2:
Regarding publishDate, EME seems to update itself to the current date automatically, though I am not sure how.

wolenetz:
For publishDate auto-update, I'll merge a fix shortly to get that working for the MSE main spec (see #130)

wolenetz:
#126 and #131 have merged.

index.html
wolenetz: With that clarification here for posterity, I agree and this change LGTM (just please regenerate the index.html after merging with gh-pages again now that #126 and #131 have landed). Thanks for fixing this, @mwatson2, and for raising this issue @annevk.

PR #129, opened 2016-07-26
Fix #126: Remove revision histories

### LINKED ISSUE #126: Consider removing the revision histories

wolenetz:
Removes the revision history section from each of the main MSE spec and
the MSE ISO-BMFF bytestream format spec; both of these documents have a
link to their github commit histories in their headers already.
Previously, keeping the revision history section maintained introduced
churn and error, and the git history is the ground truth regardless.


wolenetz:
@jdsmith3000 @mwatson2 Please take a look.
@plehegar FYI.

This should fix #126

wolenetz:
@plehegar note that publishing a PR/REC document will probably need to reference a specific branch of the github history for each of the files in that "commit history" link in the documents' headers.

jdsmith3000:
Makes sense, though we didn't do that for CR.  The link we used includes subsequent editorial changes:  https://github.com/w3c/media-source/commits/gh-pages/media-source-respec.html

It's not obvious to me what the correct link is though.  Ones I tried, like https://github.com/w3c/media-source/commits/gh-pages/media-source.html, aren't complete.

wolenetz:
Probably https://github.com/w3c/media-source/commits/v1-CR-ready/media-source-respec.html (since IIRC, that's the branch @plehegar created for the current MSE CR doc.) My comment above is that future branch for PR/REC will need adjustmend of the "commit history" links to point to the current github branch, too.

Does this PR look good to merge? If we're on board for removing the revision histories, doing this sooner rather than later will help reduce editor churn.

jdsmith3000:
Yes.  LGTM.

wolenetz:
Thanks for review.

PR #132, opened 2016-07-28
Fix #98: De-duplicate MSE and HTML51 event firings

### LINKED ISSUE #98: Steps that fire events at media elements duplicate those in HTML5.1
HTML5.1 specifies the events (e.g. `loadedmetadata`, `loadeddata`, `canplay`, `canplaythrough`)  that are to fire when the `readyState` of a media element changes, see algorithm below the definitions of the ready states in section [4.7.14.7](http://www.w3.org/TR/html51/semantics-embedded-content.html#ready-states)

The Media Source Extensions spec describes conditions that cause the `readyState` of a media element to change. Per HTML5.1, such changes imply firing relevant events at the media element.

The problem is that the Media Source Extensions spec also asks implementations to fire these events. I assume that the goal is to make it clear that changing the `readyState` of the media element also implies firing the events, but the use of normative steps effectively duplicates HTML5.1 and the spec does not say that these steps are intended to replace those in HTML5.1. In other words, an implementation that fires the `loadedmetadata` event twice would just follow the specs. Also, duplicating is error-prone, and for instance I note that the Media Source Extensions spec remains silent on the `timeupdate` and `waiting` events that get fired when the state changes from `HAVE_FUTURE_DATA` to `HAVE_CURRENT_DATA` or less (which can happen at the end of the Initialization Segment Received algorithm for example).

I suggest to leave events up to HTML5.1. This would mean dropping the steps that require events to fire, perhaps replacing them with informative notes such as "Per HTML5.1, changing `readyState` may trigger events on the media element" or something equivalent.

Sections and steps incriminated:
* [2.4.4 SourceBuffer Monitoring](http://w3c.github.io/media-source/#buffer-monitoring)
* [3.5.8 Initialization Segment Received](http://w3c.github.io/media-source/#sourcebuffer-init-segment-received), step 6.3
* [3.5.12 Coded Frame Processing](http://w3c.github.io/media-source/#sourcebuffer-coded-frame-processing), steps 2.2, 3.2, 4.2

Or is there a good reason why these steps need to be in the spec?

Note that, while this touches on normative steps, I don't think it needs to block publication as CR: the update would not change the expected behavior, but merely clarifies the intent and helps avoid misinterpretations.

wolenetz:
2 commits included (intent will be to squash when merging, but potentially ease review by keeping them separate in the PR for now):

Part 1: Fixed "removetrack" and "change" event duplication of HTML51 in
    MSE removeSourceBuffer() and "addtrack" event duplication of HTML51 in
    MSE initialization segment received algorithm by noting the expectations
    in non-normative notes rather than normative MSE text.
    Also differentiated mediatracklist vs texttracklist links into HTML51
    spec to reference the appropriate HTML51 events.

 Part 2: Fixed various HTMLMediaElement readyState transition events'
    duplication of HTML51 in MSE. For transitions previously missing any
    events in the MSE spec, added the same non-normative note that events
    may result.


wolenetz:
@jdsmith3000 @mwatson2 Please review.

wolenetz:
This htmldiff link may help review: http://services.w3.org/htmldiff?doc1=http%3A%2F%2Frawgit.com%2Fw3c%2Fmedia-source%2Fa92f091a7f3f83f487319023caae94b65c12aa11%2Findex.html&doc2=http%3A%2F%2Frawgit.com%2Fw3c%2Fmedia-source%2Fcd3756780f4a3aa79e330d1f774bff546e4a2da3%2Findex.html

wolenetz:
@jdsmith3000 @mwatson2 Friendly ping: please review.

jdsmith3000:
@wolenetz  I just reviewed the readyState change events and have two comments/questions:
- The general note stating that `HTMLMediaElement.readyState` changes may trigger events seems like it could have been said once, perhaps on the first example? Is it important to repeat it?
- One "`loadedmetadata`" event example gets explicitly described.  Is there a reason to include it in its note, while the other events aren't included?

wolenetz:
> The general note stating that HTMLMediaElement.readyState changes may trigger events seems like it could have been said once, perhaps on the first example? Is it important to repeat it?
It's not critically important to repeat it, but I couldn't find a location which stood out more than any other for putting just a single instance of this. Therefore I treated all the same, with the exception of:

> One "loadedmetadata" event example gets explicitly described. Is there a reason to include it in its note, while the other events aren't included?
I don't have a strong feeling about this one -- it was just the first one I edited when preparing this PR. In support of including the detail: many MSE apps and tests rely on this specific event, and nailing it down in the MSE spec seemed good to do, despite the potential rot/divergence if HTML changed what happens on transition of HAVE_NOTHING->HAVE_METADATA.

jdsmith3000:
I think these changes are accurate.  Each case for referencing HTMLMediaElement.readyState events seems generally commented, and as noted, specifically detailed for the "loadedmetadata" event.  

The events triggered on track changes are fully detailed though, just non-normatively.  It that sense, the duplication has been retained.  I think that's probably okay, since the precedence goes to the HTML spec.  I thought some about this, but decided your wording adds some value fitting the events into the language of the algorithms.

So, LGTM.

wolenetz:
Thanks for the review. I'll merge shortly.

PR #134, opened 2016-08-03
Fix #45: Better abstract/introduction

### LINKED ISSUE #45: Better abstract/introduction?
From
  https://lists.w3.org/Archives/Public/public-html-media/2016Jan/0000.html
[[
We would request greater use of prose to describe what the API
is doing. This is needed both to help the reader understand what is
being specified, but also to test whether the proposed API actually
meets its intended purpose.
]]

jdsmith3000:
Added a very brief text description of the API function in Introduction.


jdsmith3000:
Note:  I had to redo this PR to get it against the correct repository.

wolenetz:
LGTM. @jdsmith3000 , thanks for updating the intro. I'll merge/resolve/squash this PR shortly.

wolenetz:
To land this, I resolved merge conflicts with current head (other than publish date, I had marked the Intro as "informative", non-normative in an interim commit, triggering the conflict), and then noticed that HTML51 was no longer a normative reference as a result of @jdsmith3000's commit, so I added 39ccd1d1cf2031ee2ac56b7ab71254d844955859 to my local branch which landed both @jdsmith3000's commit as well as added HTML51 back as a normative reference.

PR #135, opened 2016-08-03
Fix #69: Remove VideoPlaybackQuality from MSE

### LINKED ISSUE #69: Move VideoPlaybackQuality object to a separate extension spec
I propose that we remove the [`VideoPlaybackQuality` object](http://w3c.github.io/media-source/#idl-def-VideoPlaybackQuality) and corresponding `getVideoPlaybackQuality()` [extension](http://w3c.github.io/media-source/#htmlvideoelement-extensions) to the `HTMLVideoElement` element from v1 to be incubated as a separate extension. (Possible outcomes include a separate Playback Performance Extensions spec or integration into to HTML 5.x.)

1. Although this functionality may be useful for adaptive streaming use cases, it is really an independent extension to `HTMLVideoElement` that is orthogonal to MSE [1]. It does not affect the core functionality of MSE and should be supported for all source types (including `.src=` ).

1. There is some question about the best way to provide information about current playback performance to applications. As one example, there is not currently a way for applications to observe changes. This capability would benefit from broader incubation, including from those that did not participate in the MSE & EME TF. If this functionality goes to REC along with the rest of MSE, any future efforts will need to deprecate part of the MSE spec or build around it.

1. One member (`totalFrameDelay`) of this object is already marked At Risk.

Moving this functionality out of the MSE spec should actually simplify the path to PR and REC by reducing the scope of the spec and removing one of the At Risk features.

Note: Moving this functionality to another spec/forum _does not_ affect what browsers implement - only the declared maturity level of the feature. Implementations that have implemented some or all of it may continue to do so, but further incubation and iteration is likely before it becomes a Recommendation. Implementation experience will be valuable input to that process.

Note: We can temporarily move these sections to a separate page within the `media-source` repo to provide an informal reference until we identify a new home.

[1] When this was added to MSE, there was no established or lightweight mechanism for small specs or extensions to existing specs. Since that time, the W3C and even HTML have adopted more iterative processes.

wolenetz:
Removes the VideoPlaybackQuality object and HTMLVideoElement extension
method getVideoPlaybackQuality(), but retains brief mention of them in a
link to the July 5, 2016 MSE Candidate Recommendation document draft in
a non-normative appendix.  Includes also an informative
reference link to the current WICG Media Playback Quality effort in that
appendix.

Also updates the "Introduction" section to be explicitly non-normative.


wolenetz:
@jdsmith3000 @mwatson2 Please take a look. I believe this aligns with the plan outlined in the current draft of the CfC for PR in the p-phme-editors thread, and it should land ASAP.

Note also that this PR includes marking the introduction section as non-normative, following the example of EME.

wolenetz:
The commit message is slightly off now that the mention in the appendix isn't in a non-normative note, rather the full section is non-normative. I'll edit the commit message, above, shortly (I think this makes it into the eventual merge.. if not, the spec itself is clear about this, and either way the text is non-normative).

wolenetz:
commit message edited, above. please review.

jdsmith3000:
LGTM.  I believe it shouldn't be merged yet however, pending the Call for Consensus response on removal of at risk features.

wolenetz:
If the CfC fails, then we can revert the merge. Since some editors will be OoO in near term for multiple weeks (IIUC), I would prefer taking the optimistic route and expect CfC passage.

paulbrucecotton:
I am okay with the Editor's plan here especially given our tight MSE [timeline](https://lists.w3.org/Archives/Public/public-html-media/2016May/0029.html) and the Editors' vacation plans.

/paulc
HME WG Chair

jdsmith3000:
I agree.  Go ahead with the merge.

wolenetz:
Merging shortly. Thanks for the review.

PR #142, opened 2016-08-09
Fix typo introduced by removing the TrackDefault feature

wolenetz:
Commit ed76bbd8f9c827771ec80df05e10983b88ed0af8 introduced a typo.
This change adds the word "equals" where it was intended to be, in the
initialization segment received algorithm's processing of audio
language, when the inband language is the 'und' BCP 47 value.


wolenetz:
This is a trivial typo fix. I'll merge this without wider review.

Issue #88, opened 2016-06-07
Clarify/remove 'stalled' and 'progress' events for MediaSource


chcunningham:
Note on priority: Happy to see any spec change pushed to MSE vNext. v1 is nearing cut date and IMO this isn't critical.

'Stalled' makes good sense for non-MediaSource playbacks where the UA is responsible for the fetching the media. For MSE, the web app does the fetching and MSE is unaware that the fetch is being made and cannot know whether its progressing or stalled. The MSE spec states that "the download" or "bytes received" (in resource fetch algorithm) refer to data passed in via appendBuffer() and appendStream()". With this model, it is not possible for MSE to experience a stall (appends don't stall). 

MSE spec makes no special mention of 'stalled' and when to emit (if ever) is ambiguous given the lines about "the download" and "bytes received" above. At the moment Chrome raises stalled if an MSE app goes 3 seconds without appending. This [leads to confusion](crbug.com/517240). I propose that we should not emit stalled for MSE (similar to the decision to not emit suspend for MSE in [Issue 24](https://github.com/w3c/media-source/issues/24)).

When to emit 'progress' is also not clear. Raising this during an append is not as confusing as Chrome's behavior for 'stalled', but one could argue that raising progress is still confusing (given that app control the fetch) and also not useful for app developers. Apps know when they've called append and they can clearly know the outcome/completion of that append by other events/states (e.g. updateend event, SourceBuffer.buffered). I propose that we should also not bother to emit progress.

This is also a call for app developers to point out MSE use cases for these events that I may have overlooked. 

One argument that I've heard is that some players may have UI that relies on these events dating back pre-MSE implementation. If this is a real concern then we have to decide on an implementation for these events that is meaningful. Showing a stall indicator because no append has been made in the last 3 seconds is not a great UI. Apps may append less frequently than that and still deliver a perfect playback experience. Additionally, apps would be better served basing any network UI on the network fetch they make prior-to appending to MSE.

chcunningham:
Sorry, closed by mistake

chcunningham:
reference: #24 
fyi: @wolenetz @foolip @karlt @mounirlamouri

wolenetz:
Thanks for filing this, @chcunningham.
I suppose 'stalled' and 'progress' might make more sense to retain for MSE *if* we retained appendStream(). However, for V1 MSE, appendStream() is currently marked at-risk.

[...]

PR #143, opened 2016-08-09
Fix #99: Update the attachment references to lead to HTML51 local mode

### LINKED ISSUE #99: The fetch step in HTML5.1's resource fetch algorithm no longer exists
In [2.4.1 Attaching to a media element](https://w3c.github.io/media-source/#mediasource-attach), the Media Source Extensions specification plugs itself into the "Perform a potentially CORS-enabled fetch" step of the [resource fetch algorithm](https://www.w3.org/TR/html51/semantics-embedded-content.html#resource-fetch-algorithm), defined in HTML5.1.

This algorithm has been updated in HTML5.1 and that step no longer exists.

Actually, I think that algorithm got updated precisely to cover [media provider objects](https://w3c.github.io/html/semantics-embedded-content.html#media-provider-object) à la `MediaSource`, although the expected attachment method is rather via `srcObject`, which seems planned for VNext (#28).

Perhaps you can plug after the first step in this algorithm to force `mode` to `local`, reset the `delaying-the-load-event-flag`, etc. Step 4. in "Attaching to a media element" would then not be needed anymore. I may have missed something though.

Also note that steps à la "If the media data cannot be fetched at all, ..." are now defined in a separate [media data processing steps list](https://w3c.github.io/html/semantics-embedded-content.html#media-data-processing-steps-list) algorithm and are no longer in the resource fetch algorithm. References should be updated accordingly in the [End of stream algorithm](https://w3c.github.io/media-source/#end-of-stream-algorithm).

wolenetz:
Note that the old "stalled" and "progress" event logic is allowed (but not required)
in a non-normative note (even though the logic which does such events is in the
now-skipped remote mode section) since that particular behavior change
would have been substantive, IMHO. We already have a VNext MSE spec
bug #88 which will eventually clarify behavior around 'stalled' and
'progress' events.

edit: included "progress" in this msg to inform eventual squash commit msg, since 3c44929 now also mentions 'progress' events.


wolenetz:
@jdsmith3000 @mwatson2 @tidoust: I believe you are all out of the office this week. This PR follows the route outlined by each of @tidoust, @mwatson2, and @foolip in comments in #99.

I would like to merge this ASAP to meet MSE timeline. Given that I believe this follows the editors' agreed-upon implementation route, I would like at least one other person to look this over before I merge it. @foolip or @paulbrucecotton, would you please take a look and see if this raises any concerns? I kept the substantive changes out of this by noting the same in non-normative notes (specifically, 'stalled' MAY still occur; and srcObject MAY not reflect the src-attribute (or <source> child src-attribute) attached MediaSource object).
Barring any comments otherwise, I'll plan to merge this Wednesday Aug 10 before noon PDT.

wolenetz:
Update link for reviewers (including 7fa73b0): http://services.w3.org/htmldiff?doc1=http%3A%2F%2Frawgit.com%2Fw3c%2Fmedia-source%2Fba1a409a311ad642dd787adcb1535eda3c95e268%2Findex.html&doc2=http%3A%2F%2Frawgit.com%2Fw3c%2Fmedia-source%2F7fa73b0b802523aab078c2f92bdbe1b6db81755c%2Findex.html

paulbrucecotton:
>I would like to merge this ASAP to meet MSE timeline. 

@wolenetz - As much as I appreciate your attempts to close out the remaining MSE GitHub issues ASAP, maybe a better plan would be for you to leave this PR open for a longer review and for you to work on the outstanding comments and problems with the MSE test suite on WPT?  The test suite and test results are as much the "long pole" to getting to PR as the GitHub issues are.

/paulc

index.html
```diff
@@ -1782,7 +1786,7 @@ <h4 id="h-mediasource-detach" resource="#h-mediasource-detach"><span property="x
             <a href="https://www.w3.org/TR/html51/webappapis.html#queuing">Queue a task</a> to <a href="https://www.w3.org/TR/html51/infrastructure.html#fire">fire a simple event</a> named <code><a href="#dom-evt-sourceclose">sourceclose</a></code> at the <a href="#idl-def-mediasource" class="internalDFN" data-link-type="dfn"><code>MediaSource</code></a>.</li>
         </ol>
         <div class="note">
-          <div class="note-title marker" aria-level="5" role="heading" id="h-note20"><span>Note</span></div>
+          <div class="note-title marker" aria-level="5" role="heading" id="h-note21"><span>Note</span></div>
```
foolip: I see no internal links to #h-note*, so I guess these are linked to from somewhere else? Aren't they meant to be stable? If nobody is actually expected to link to these, I'd just drop the ids to make diffs like this smaller :)

foolip: Oh, index.html is generated, media-source-respec.html is the input. Too bad they show up in the other order in diffs :)

media-source-respec.html
```diff
@@ -653,29 +653,30 @@
...
-          <p>If the <a def-id="resource-fetch-algorithm"></a> absolute URL matches the MediaSource object URL, ignore any <a def-id="preload"></a> attribute of the media element, skip any optional steps for when <a def-id="preload"></a> equals <a def-id="preload-none"></a>, and run the following steps right before the <a def-id="perform-potentially-cors-enabled-fetch"></a>
-            step in the <a def-id="resource-fetch-algorithm"></a>.</p>
 
+          <p>If the <a def-id="resource-fetch-algorithm"></a> was invoked with a media provider object or a URL record whose object is a media provider object that is a MediaSource object, then let mode be local, skip the first step in the <a def-id="resource-fetch-algorithm"></a> which may otherwise set mode to remote, and run the following steps at the beginning of the <a def-id="Otherwise-mode-is-local"></a> section of the <a def-id="resource-fetch-algorithm"></a>.
```
foolip: Instead of "URL record whose object is a media provider object that is a MediaSource object" you can simply say "URL record whose object is a MediaSource object", since "media provider object" isn't a thing in itself, but just a shorthand for a  MediaStream, MediaSource or Blob.

index.html
wolenetz: Yes, the regeneration upon note insertion/removal makes for a bunch of lines in the index.html diff. I suppose the link tag is generated in case someone might wish to make a link into a document revision to a particular note, though those numbers can obviously change significantly across revisions.

media-source-respec.html
wolenetz: The "that is a MediaSource object" phrase also applies to the "invoked with a media provider object" portion. I've updated the wording in 9310bf8 to be more clear (using your suggestion, plus clarifying the "invoked with a media provider object **that is a MediaSource object**...")

wolenetz:
9310bf8 addresses all comments so far.
Updated htmldiff link to assist review: http://services.w3.org/htmldiff?doc1=http%3A%2F%2Frawgit.com%2Fw3c%2Fmedia-source%2Fba1a409a311ad642dd787adcb1535eda3c95e268%2Findex.html&doc2=http%3A%2F%2Frawgit.com%2Fw3c%2Fmedia-source%2F9310bf8b095dbe654f313de4a29ee1859d26f9a7%2Findex.html

media-source-respec.html
```diff
@@ -653,29 +653,30 @@
...
+          <p class="note">The <a def-id="resource-fetch-algorithm"></a>[[HTML51]]'s first step is expected to eventually align with selecting local mode for URL records whose objects are media provider objects. The intent is that if the HTMLMediaElement's <code>src</code> attribute or selected child <code>&lt;source&gt;</code>'s <code>src</code> attribute is an absolute URL matching a <a def-id="MediaSource-object-URL"></a> when the respective <code>src</code> attribute was last changed, then that MediaSource object is used as the media provider object and current media resource in the local mode logic in the <a def-id="resource-fetch-algorithm"></a>. This also means that the remote mode logic that includes observance of any preload attribute is skipped when a MediaSource object is attached. Even with that eventual change to [[HTML51]], the execution of the following steps at the beginning of the local mode logic is still required when the current media resource is a MediaSource object.</p>
```
foolip: Two issues on the same source line.

A bit odd with the spec ref is between "resource fetch algorithm" and the trailing 's, maybe skip the ref entirely or put if after?

Maybe "`blob:` URL" instead of "absolute URL"?

media-source-respec.html
```diff
@@ -653,29 +653,30 @@
...
+                    <li>Text in the <a def-id="resource-fetch-algorithm"></a> or the <a def-id="media-data-processing-steps-list"></a> that refers to "the download", "bytes received", or "whenever new data for the current media resource becomes available" refers to data passed in via <a def-id="appendBuffer"></a>.</li>
+                    <li>References to HTTP in the <a def-id="resource-fetch-algorithm"></a> and the <a def-id="media-data-processing-steps-list"></a> do not apply because the HTMLMediaElement does not fetch media data via HTTP when a <a>MediaSource</a> is attached.</li>
                   </ol>
                 </li>
               </ol>
             </dd>
           </dl>
+          <p class="note">An attached MediaSource does not use the remote mode steps in the <a def-id="resource-fetch-algorithm"></a>, so the media element will not fire "suspend" events. User agents MAY still fire "stalled" events as these [[HTML51]] references changed after implementations of this specification stabilized.</p>
```
foolip: The spec lawyer in me would say that having the MAY in a note is weird, as it's normative language in a non-normative context. Can it be made clear what the hoped-for eventual behavior is, so that tests for that can be added to web-platform-tests (or somewhere) without controversy?

media-source-respec.html
```diff
@@ -653,29 +653,30 @@
...
+          <p class="note">When a MediaSource is attached via <code>src</code> attribute or selected child <code>&lt;source&gt;</code>'s <code>src</code> attribute set to a <a def-id="MediaSource-object-URL"></a>, then the <code>srcObject</code> attribute of the HTMLMediaElement MAY not reflect this as these [[HTML51]] references changed after implementations of this specification stabilized.</p>
```
foolip: Hmm, not sure I understand this note. Even the way this is intended to work in WHATWG HTML, using a `blob:` URL in the `src` attribute won't cause the internal object to show up in `srcObject`. I expressed a similar idea in https://github.com/whatwg/html/issues/295#issuecomment-152169762 but it didn't turn out that way in the end.

foolip:
OK, I've finished reviewing the whole change, on the assumption that it'll be squashed into a single commit so I didn't pay attention to the later commit messages. LGTM, I think it does the job, and since the difference between this sometimes taking the remote branch is observable, it's important to get all implementations aligned.

(And again, I'm sorry that integrating with HTML is such a pain for you, but I'll do what I can to make it less terrible.)

media-source-respec.html
wolenetz: Oh. That's not quite how either mwatson2@ (in https://github.com/w3c/media-source/issues/99#issuecomment-234029419) or I understood this, but rereading the current whatwg spec (https://html.spec.whatwg.org/multipage/embedded-content.html#dom-media-srcobject) shows that assigned media provider object isn't changed by the resource fetch algorithm. Could it be? The idea is that it would rock if devtools users could query a media element for its srcObject to aid debugging of MediaSource (or MediaStream, etc) without modifying the original page to expose the MediaSource object some other way.

media-source-respec.html
wolenetz: We have several occurrences of such wording in other notes, but I get your point. I'll fix this one to clarify that the hoped-for behavior would be to **not** emit 'stalled' for MSE-attached media elements, since that logic is contained in the 'remote' branch. I'll expand this to also include similar hope for no 'progress' events in MSE-attached media elements.

media-source-respec.html
wolenetz: sgtm. I'll fix both. I used "absolute URL" since that's what the current w3c html 5.1 terminology is, but blob: is even more specific, and this is already in a non-normative note describing the real intent.

media-source-respec.html
wolenetz: For now, I'll simply remove this note. If HTML5.1 changes behavior around this, that would be outside the scope of MSE :).

wolenetz:
3c44929 addressess all comments so far. Updated htmldiff link to assist review is: http://services.w3.org/htmldiff?doc1=http%3A%2F%2Frawgit.com%2Fw3c%2Fmedia-source%2Fba1a409a311ad642dd787adcb1535eda3c95e268%2Findex.html&doc2=http%3A%2F%2Frawgit.com%2Fw3c%2Fmedia-source%2F3c449292ab57363913384b4a00593c47f64e4cdf%2Findex.html

foolip:
I agree that having `srcObject` automatically reflect the object inside a `blob:` URL assigned to the `src` attribute would be neat and at least occasionally useful. Simply extracting and exposing the object in `srcObject` at the beginning of the resource selection algorithm is probably low risk, better than the `src` attribute idea I had. The weird bit might be what the `currentSrc` attribute should then return. These details are probably not insurmountable, if you think something like that's a good idea and would implement if spec'd, feel free to file an HTML spec issue or send a PR of how you'd like it work.

(Top-level since it'd otherwise be in collapsed comments.)

wolenetz:
Note, this and PR #146 modify similar region of the spec, so merge conflict resolution will be necessary when they're landed.

foolip:
This PR LGTM, can it be merged so reviewing the other one is cleaner?

index.html
```diff
@@ -1733,28 +1733,36 @@ <h3 id="h-mediasource-algorithms" resource="#h-mediasource-algorithms"><span pro
...
 
+        <p>If the <a href="https://www.w3.org/TR/html51/semantics-embedded-content.html#resource-fetch-algorithm">resource fetch algorithm</a> was invoked with a media provider object that is a MediaSource object or a URL record whose object is a MediaSource object, then let mode be local, skip the first step in the <a href="https://www.w3.org/TR/html51/semantics-embedded-content.html#resource-fetch-algorithm">resource fetch algorithm</a> which may otherwise set mode to remote, and run the following steps at the beginning of the <span>"<i>Otherwise (mode is local)</i>"</span> section of the <a href="https://www.w3.org/TR/html51/semantics-embedded-content.html#resource-fetch-algorithm">resource fetch algorithm</a>.
```
jdsmith3000: This follows the model of the previous section contents, but feels awkward.  The goal is to set the mode to local, skip the first resource fetch algorithm step (which could set mode to remote), add some steps that are MediaSource specific, and modify the "Otherwise (mode is local)" steps to fit.

Since it follows the previous structure here, it's probably okay.  I'd be tempted to say something more like:
`If...then let mode be local, skip the first step in the resource fetch algorithm (which may otherwise set the mode to remote, and add the steps and clarifications below to the "Otherwise (mode is local)" section of the resource fetch algorithm.

This might seem like wordsmithing.  I'm reacting a bit to the "before" aspect of the steps, when they are partly before, and partly alterations of the steps themselves.

media-source-respec.html
```diff
@@ -653,29 +653,29 @@
...
           <dl class="switch">
             <dt>If <a def-id="readyState"></a> is NOT set to <a def-id="closed"></a></dt>
-            <dd>Run the <a def-id="media-data-cannot-be-fetched"></a> steps of the <a def-id="resource-fetch-algorithm"></a>.</dd>
+            <dd>Run the <a def-id="media-data-cannot-be-fetched"></a> steps of the <a def-id="resource-fetch-algorithm"></a>'s <a def-id="media-data-processing-steps-list"></a>.</dd>
```
jdsmith3000: This addition LGTM.

media-source-respec.html
```diff
@@ -653,29 +653,29 @@
...
-          <p>If the <a def-id="resource-fetch-algorithm"></a> absolute URL matches the MediaSource object URL, ignore any <a def-id="preload"></a> attribute of the media element, skip any optional steps for when <a def-id="preload"></a> equals <a def-id="preload-none"></a>, and run the following steps right before the <a def-id="perform-potentially-cors-enabled-fetch"></a>
-            step in the <a def-id="resource-fetch-algorithm"></a>.</p>
 
+          <p>If the <a def-id="resource-fetch-algorithm"></a> was invoked with a media provider object that is a MediaSource object or a URL record whose object is a MediaSource object, then let mode be local, skip the first step in the <a def-id="resource-fetch-algorithm"></a> which may otherwise set mode to remote, and run the following steps at the beginning of the <a def-id="Otherwise-mode-is-local"></a> section of the <a def-id="resource-fetch-algorithm"></a>.
```
jdsmith3000: Moving comment to respec file:

This follows the model of the previous section contents, but feels awkward. The goal is to set the mode to local, skip the first resource fetch algorithm step (which could set mode to remote), add some steps that are MediaSource specific, and modify the "Otherwise (mode is local)" steps to fit.

Since it follows the previous structure here, it's probably okay. I'd be tempted to say something more like:

`If...then let mode be local, skip the first step in the resource fetch algorithm (which may otherwise set the mode to remote, and add the steps and clarifications below to the "Otherwise (mode is local)" section of the resource fetch algorithm.`

This might seem like wordsmithing. I'm reacting a bit to the "before" aspect of the steps, when they are partly before, and partly alterations of the steps themselves.

jdsmith3000:
I made a minor comment about how the added/clarified steps are spliced in.  Otherwise, the change PR LGTM as well.

media-source-respec.html
wolenetz: @jdsmith3000 your suggested slight reword seems benign and I'll update and merge this PR shortly, accordingly. The previous "beginning" wording was there since the intent is to run the attachment ASAP within the mode-is-local branch, but there are no other "steps" directly within that w3c html51 mode-is-local piece, so "beginning" is less meaningful.

PR #146, opened 2016-08-11
Fix #104: Clarify that attachment is asynchronous

### LINKED ISSUE #104: Consider making MSE attachment asynchronous
@jdsmith3000, @mwatson2  : what do you think? Is there any real restriction on requiring the attachment be synchronous? Would allowing attachment to be done asynchronously constitute a substantive change?

@foolip This follows our discussions on https://bugs.chromium.org/p/chromium/issues/detail?id=615481#c14

wolenetz:
Also clarify that the steps which transition the MediaSource from
"closed" to "open" readyState during attachment may be delayed so the
implementation can ensure the MediaSource object is ready for use.


wolenetz:
@foolip @paulbrucecotton @jdsmith3000 @mwatson2 Please take a look.
Since this is in a non-normative note, and doesn't restrict existing behavior, though allows some previously less-clarified-delay-during-attachment behavior in some UAs, I believe this is likely non-substantive, but helpful to include in the spec. IIUC, @jdsmith3000 and @mwatson2 are OoO through next week, so I'll depend on @foolip and @paulbrucecotton on this for review and guidance around substantiveness.

wolenetz:
@jdsmith3000 @mwatson2 Please review this ASAP per the editors' call this morning. Thank you.

media-source-respec.html
```diff
@@ -660,6 +660,7 @@
...
           <dl class="switch">
+            <p class="note">Relative to the action which triggered the media element's resource selection algorithm, these steps are already asynchronous. The resource fetch algorithm is run after the task that invoked the resource selection algorithm is allowed to continue and a stable state is reached. Implementations may delay the steps in the Otherwise clause, below, until the MediaSource object is ready for use.</p>
```
jdsmith3000: nit:  I suggest deleting "already" from "already asynchronous".

jdsmith3000: nit:  I believe "Otherwise" should be italicized and in quotes to be consistent with our conventions.

jdsmith3000:
Minor nits.  I like your phrasing on `Implementations may delay the steps in the Otherwise clause, below, until the MediaSource object is ready for use`.  This warns of timing considerations nicely.

Outside of nits, LGTM.

wolenetz:
@jdsmith3000 Thanks for review. I'll address both of the nits and merge this shortly.

PR #148, opened 2016-08-17
Fix #124: Plan C: Allow mimicking of partial audio frame rendering

### LINKED ISSUE #124: Confusing duration change consequences and inconsistent outcome for user agents.
In the updated spec, we have:
http://w3c.github.io/media-source/#duration-change-algorithm

> "5. If a user agent is unable to partially render audio frames or text cues that start before and end after the duration, then run the following steps:
> Note
> 
> This condition can occur because the coded frame removal algorithm preserves audio frames and text cues that start before and end after the duration.
> 
>     1. Update new duration to the highest end time reported by the buffered attribute across all SourceBuffer objects in sourceBuffers.
>     2. Update duration to new duration.
> "

I'll explain the issue with an example. Assuming we have a video sourcebuffer with the following 3 samples:
[0,2)[2,4)[4,6)

if we call remove with the arguments (3, 6), the resulting sourcebuffer will now contain:
[0,2)[2,4)

as a sample is only removed as : "Remove all media data, from this track buffer, that contain starting timestamps greater than or equal to start and less than the remove end timestamp." (from Coded Frame Removal)

The sample [2,4) isn't truncated.

When setting duration to 3 however, if the usage agent is able to partially render audio frames, then the resulting duration will stay 3.
However, for a user agent enable to partially render audio frames, the duration will now be 

> "Update new duration to the highest end time reported by the buffered attribute across all SourceBuffer objects in sourceBuffers."

so duration for those user agents will be 4 instead.

Both type of user agents will have a buffered range of [0,4), yet one has a duration of 3 and the other 4 (which is even more confusing as we do not have to worry about audio frames here)

I don't believe that there is anything in the spec, indicating that playback would have to stop once currentTime reaches duration; on the contrary:
https://w3c.github.io/html/semantics-embedded-content.html#offsets-into-the-media-resource

> "The duration attribute must return the time of the end of the media resource, in seconds, on the media timeline. ".

When played, that last sample will update the duration, in which case once playback has ended we would have a final duration = 4.

As such, I believe that 

> "5.If a user agent is unable to partially render audio frames or text cues that start before and end after the duration, then run the following steps" should be removed to ensure consistency across user agents. Step 4 "Update duration to new duration."

 would need to be removed.

Additionally, as side effects of those steps that change the duration to what was originally set.
if you do:
mediaSource.duration = 3;
mediaSource.duration = 3;

as in both cases the duration was actually set to 4, the first step of the duration change algorithm 

> "1. If the current value of duration is equal to new duration, then return."

will always be false; and the durationchange event will be fired twice (this issue arised in the web-platform-test media-source/mediasource-duration.html which specifically check that durationchange event is not fired when setting the duration to the same value multiple times in a row).

As a workaround for this issue, I suggest that in the final step, prior "Update the media duration to new duration and run the HTMLMediaElement duration change algorithm."

we do: "If the value of old duration is equal to new duration, then return."
This will ensure that no durationchange event will be fired.

wolenetz:
This follows "Plan C" laid out in
https://github.com/w3c/media-source/issues/124#issuecomment-240530716

### DIFF
--- a/index.html
+++ b/index.html
@@ -872,7 +872,7 @@ github.com style (c) Vasily Polovnyov <vast@whiteants.net>
       <a class="logo" href="https://www.w3.org/"><img width="72" height="48" src="https://www.w3.org/StyleSheets/TR/2016/logos/W3C" alt="W3C"></a>
     </p>
     <h1 class="title p-name" id="title" property="dcterms:title">Media Source Extensions</h1>
-    <h2 id="w3c-editor-s-draft-16-august-2016"><abbr title="World Wide Web Consortium">W3C</abbr> Editor's Draft <time property="dcterms:issued" class="dt-published" datetime="2016-08-16">16 August 2016</time></h2>
+    <h2 id="w3c-editor-s-draft-17-august-2016"><abbr title="World Wide Web Consortium">W3C</abbr> Editor's Draft <time property="dcterms:issued" class="dt-published" datetime="2016-08-17">17 August 2016</time></h2>
     <dl>
       <dt>This version:</dt>
       <dd><a class="u-url" href="http://w3c.github.io/media-source/">http://w3c.github.io/media-source/</a></dd>
@@ -1981,15 +1981,11 @@ interface <span class="idlInterfaceID">MediaSource</span> : <span class="idlSupe
             <p class="">Duration reductions that would truncate currently buffered media are disallowed. When truncation is necessary, use <code><a href="#dom-sourcebuffer-remove">remove()</a></code> to reduce the buffered range before updating <code><a href="#dom-mediasource-duration">duration</a></code>.</p>
           </div>
           <li>Update <code><a href="#dom-mediasource-duration">duration</a></code> to <var>new duration</var>.</li>
-          <li>If a user agent is unable to partially render audio frames or text cues that start before and end after the <code><a href="#dom-mediasource-duration">duration</a></code>, then run the following steps:</li>
           <div class="note">
             <div class="note-title marker" aria-level="5" role="heading" id="h-note31"><span>Note</span></div>
-            <p class="">This condition can occur because the <a href="#sourcebuffer-coded-frame-removal">coded frame removal algorithm</a> preserves audio frames and text cues that start before and end after the <code><a href="#dom-mediasource-duration">duration</a></code>.</p>
+            <p class="">Partial coded frames can occur because both this algorithm and the <a href="#sourcebuffer-coded-frame-removal">coded frame removal algorithm</a> preserve audio frames and text cues that start before and end after the <code><a href="#dom-mediasource-duration">duration</a></code>. Implementations that do not support partial audio coded frame rendering may play decoded audio data beyond duration for the entirety of the presentation interval of audio coded frames that start before and end after the <code><a href="#dom-mediasource-duration">duration</a></code>. HTMLMediaElement currentTime and buffered range end times should not exceed duration, even with this allowance.
+            </p>
           </div>
-          <ol>
-            <li>Update <var>new duration</var> to the highest end time reported by the <code><a href="#dom-sourcebuffer-buffered">buffered</a></code> attribute across all <a href="#idl-def-sourcebuffer" class="internalDFN" data-link-type="dfn"><code>SourceBuffer</code></a> objects in <code><a href="#dom-mediasource-sourcebuffers">sourceBuffers</a></code>.</li>
-            <li>Update <code><a href="#dom-mediasource-duration">duration</a></code> to <var>new duration</var>.</li>
-          </ol>
           <li>Update the <code><a href="https://www.w3.org/TR/html51/semantics-embedded-content.html#dom-htmlmediaelement-duration">media duration</a></code> to <var>new duration</var> and run the <a href="https://www.w3.org/TR/html51/semantics-embedded-content.html#durationChange">HTMLMediaElement duration change algorithm</a>.</li>
         </ol>
       </section>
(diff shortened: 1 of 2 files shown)


wolenetz:
@jdsmith3000 @mwatson2 @jyavenard Please take a look.

jdsmith3000:
This PR asserts that it is preferable to introduce a new behavior (holding duration, but optionally allowing playback on data that goes outside of it), rather than changing the spec to allow just one of two previously spec'd solutions.  The advantage of the proposed change is that duration is not modified by the UA to reflect the data endpoints, but the disadvantage is that playback will vary from one UA to the other, without no change that may be detected by the app.  Given that:

1.  What is your rationale for saying that extending duration to match the endpoints presents an interop issue when it was always allowed by the spec?
2.  Why do you believe that the new hybrid behavior of not extending duration, but allowing playback beyond it's limits poses no interoperability risk?

wolenetz:
@jdsmith3000 in response to https://github.com/w3c/media-source/pull/148#issuecomment-240587868 

> 1. What is your rationale for saying that extending duration to match the endpoints presents an interop issue when it was always allowed by the spec?

a) As @jyavenard pointed out in https://github.com/w3c/media-source/issues/124#issuecomment-239591586, the hybrid behavior in Chrome was potentially becoming de-facto standard unfortunately out-of-line with the spec. This PR attempts to align the spec with the hybrid behavior.
b) Chrome has been using a hybrid approach which precludes extending duration to the highest coded frame end time. I don't have data for but suspect that extremely long text cue durations could cause the resulting duration to be strangely large, and remove(desired_duration, current_duration) would *not* afford an app opportunity to easily rectify this situation without something like this plan C approach.

> 2. Why do you believe that the new hybrid behavior of not extending duration, but allowing playback beyond it's limits poses no interoperability risk?

I didn't assert there was no interoperability risk (see https://github.com/w3c/media-source/issues/124#issuecomment-240549687). Rather, less interoperability risk, since at least there would be a single app-visible expectation (unless the app has hard dependencies on currentTime polling compared to "time until 'ended' event receipt", which is already asynchronous so is already not a good thing for an app to depend on; also an app might slave webaudio to the media element and observe the differences in number of audio samples afforded by the allowance in plan C. But such variance is documented as part of plan C.) Regarding app-risk of the hybrid approach formalized in this PR, I'm not aware of any major concerns Chrome users have had with the current hybrid implementation.

jdsmith3000:
How does the Chrome hybrid behavior handle the case where text cue durations are extremely long?  I assume it would continue rendering the cue until the data was exhausted, even though duration had been exceeded.  

I don't read @jyavenard's comment as saying the Chrome behavior was becoming defacto standard.  He said that FF plays on to exhaust all data, but comments also that this aligns with duration, which had previously been adjusted to match:

> Seeing that the duration has already been adjusted to be exactly at the end of the last frame, currentTime when playback ends is always exactly equal to the set duration, and no durationchange event will be fired.

This is a fully defined behavior in the CR spec, and to me has merit, since duration aligns with the buffered data boundaries.  In fact, the change in duration signals that buffered data would play beyond the original duration.

I intend to consult with a Media dev here on this, but don't believe that can be done until tomorrow morning.

wolenetz:
OK. With a bit of rethinking about potential impact here, and trusting that Gecko and Edge haven't incurred web author complaints akin to "my app's broken because it expects the duration to be X when it sets the duration to X", I think plan B is more fair to all of us to use. I'll prepare a separate PR to see what that looks like.

If a web author truly desires that the duration of the buffered arbitrary media be X (where X < arbitrary media's highest buffered end time), plan B will not guarantee that. Nor will a potential polyfill which appends overlapping media for all tracks into the timeline at time X and then calls remove(X,duration) followed by setting duration to X. Such polyfill would work for audio and text tracks, but the overlapped video frame is left untouched in the track buffer and so would retain its original duration.

Could we therefore expand Plan B to include support for potential polyfill to obtain precise duration if the app truly desires that? (For instance, include a video splice frame algorithm which just adjusts the overlapped frame duration downwards to make it appear as if it were originally immediately adjacent to the overlapping frame (much like the text splice algorithm).) Or is such polyfill support need mitigated by apps' ability to parse and mux their content client- or server-side appropriately to achieve whatever precise duration they desire?

I'd prefer simpler change at this point, so plan B without polyfill support, but definitely with a big warning in the MediaSource.duration spec for the setter to web authors that the setter may result in duration being adjusted to some higher value than it was set to as part of the synchronous duration property set operation. If we go this route, perhaps a VNext issue to add a video splice frame algorithm (to support such polyfill) would be appropriate.

Thanks for your patience with me while I worked through this -- I'm trying to keep our API's usability as the focus throughout.

jdsmith3000:
I don't quite follow.  Does the polyfill splice partial frame data to even out track durations?  If so, how might that allow an app to exert more precise control over the duration value?

jyavenard:
I find this approach cumbersome and rather non intuitive... 

Playing data, yet suspending currentTime to represent that it's playing.
Worse, I imaging that it will allow ended to be fired while playback is still ongoing.
What if the media source also goes back into readyState = "open", should the currentTime still be capped?

I don't see what breakage the change discussed in #124. Firefox is already doing it as per spec (following the condition that partial frames can't be rendered)... so player not following the spec, would already be broken with firefox.

Alternatively, impose that all UA must support partial frames.

wolenetz:
@jdsmith3000  (https://github.com/w3c/media-source/pull/148#issuecomment-240602250):

> Does the polyfill splice partial frame data to even out track durations? If so, how might that allow an app to exert more precise control over the duration value?

Yes, a theoretical polyfill could use crafted overlap-appends to obtain a more precise end time. overlap-append, followed by remove() and setting duration would be the general approach. I don't think that polyfill is totally blocked if we go with a plan B approach, so let's table any missing support for it for VNext if ever.

wolenetz:
@jyavenard (https://github.com/w3c/media-source/pull/148#issuecomment-240741688)

I agree. One item that the plan B approach will need will be to address the precise language from:

> the highest end time reported by the buffered attribute across all SourceBuffer objects in sourceBuffers

to:
"the highest track buffer ranges end time across all track buffers for all sourceBuffer objects in sourceBuffers"

This addresses the current spec still affording/requirinq partial frame rendering support for muxed SourceBuffers by removing partial frame rendering impliciations even in those cases. I'll produce a separate PR for this "plan B" shortly.

jdsmith3000:
Sounds good.  

FYI:  I have confirmed that Edge currently does not support partial frame rendering and we have no wish to do so.  Our current (what what we feel is desirable) behavior then should match what I believe we've agreed upon here.

wolenetz:
It looks like we have consensus that Plan C is no-go for V1, and plan B is the way forward. I'll produce a separate PR for plan B.

PR #150, opened 2016-08-17
Fix #149: Exclude text track buffers from SourceBuffer.buffered logic

### LINKED ISSUE #149: Clarify SourceBuffer.buffered logic for in-band text tracks
Text track cues are not necessarily continuous, but their lack of continuity should not cause media playback stall.

Since the SourceBuffer.buffered attribute is the root of the data used by the MSE-extended HTMLMediaElement.buffered to determine which regions of the timeline can playback without stall, I believe SourceBuffer.buffered's spec text needs to make special note that discontinuous text tracks do not, by themselves, trigger buffered range discontinuity for the media.

I believe the best way to accomplish this in the spec would be to indicate in the buffered logic something like:

Instead of:
"For each track buffer managed by this SourceBuffer, run the following steps:", replace with:
"For each audio and video track buffer managed by this...." along with a non-normative note explaining why text tracks are left out of the calculation.

@jdsmith3000 Please comment ASAP; I believe this is V1Editorial. FWIW, Chromium would not need a change to match this proposed text.

wolenetz:


wolenetz:
@jdsmith3000 Please take a look.

jdsmith3000:
Makes sense.  LGTM.

PR #151, opened 2016-08-17
Fix #147: Clarify that seeks must complete to all positions within hme-buffered

### LINKED ISSUE #147: Seeking in the buffered range should always succeed.
From: https://w3c.github.io/media-source/index.html#mediasource-seeking
describe the steps on how to seek with MSE
"If one or more of the objects in activeSourceBuffers is missing media segments for the new playback position
  If the HTMLMediaElement.readyState attribute is greater than HAVE_METADATA, then set the HTMLMediaElement.readyState attribute to HAVE_METADATA.
"

One can implicitly assume that seeking in the reported buffered range will always succeed.
in particular as we have "The web application can use buffered to determine what the media element needs to resume playback."

When the mediasource readyState is "open", the buffered range only reports where we actually have data.
However, if the readyState is "ended" (https://w3c.github.io/media-source/index.html#widl-SourceBuffer-buffered) "If readyState is "ended", then set the end time on the last range in track ranges to highest end time."

So let's assume that we have a media element that is made of an audio and video tracks.
The audio track has a buffered range of [5-10] and the video track has [5-20].
When in "ended" readyState, the buffered attribute reported will as such be [5-20]" ; however we do not have audio data in the [10-20] range; so according to the seeking definition, the media element readyState should become HAVE_METADATA and seek can't succeed.

As such two possibilities:
1- Have buffered always return the buffered range intersection of all tracks, regardless of the readyState value
2- Specify in the seeking steps, that if the reported buffered range contains the position, seeking must complete (so the UA when completing the seek would then play whatever left data it has without stalling to the end of the buffered range)

I favour option 2.

wolenetz:
This corresponds to option 2 discussed in issue #147.


wolenetz:
@jdsmith3000 Please take a look. Thanks.

media-source-respec.html
```diff
@@ -705,7 +705,8 @@
           <h4>Seeking</h4>
           <p>Run the following steps as part of the "<i>Wait until the user agent has established whether or not the media data for the new playback position is available, and, if it is, until it has decoded enough data to play back that position"</i> step of the <a def-id="hme-seek-algorithm"></a>:</p>
           <ol>
-            <li>The media element looks for <a def-id="media-segments"></a> containing the <var>new playback position</var> in each <a>SourceBuffer</a> object in <a def-id="activeSourceBuffers"></a>.
+            <li>The media element looks for <a def-id="media-segments"></a> containing the <var>new playback position</var> in each <a>SourceBuffer</a> object in <a def-id="activeSourceBuffers"></a>. Any position within a <a def-id="timerange"></a> in the current value of the <a def-id="hme-buffered"></a> attribute MUST have a buffered media segment for that position.
```
jdsmith3000: This doesn't read like a step.  I believe your intent is to add a step somewhat like this:

`If the readyState attribute is "ended" and the new playback position is within a TimeRange currently in HTMLMediaElement.buffered, then the seek must complete even if one or more currently selected track buffers have end times beyond the coded frame presentation interval.`

I feel like the contents of the note that follows belong in the step itself.

wolenetz: @jdsmith3000 I wasn't proposing a new step, rather clarifying the existing one. Are you proposing a conditional sub-step like:
Existing step text + "If readyState==ended, then:
1)"[ something like your text ]

I agree the text in the note that follows should be normative, so putting in outside of the note would be better. I'll proceed with a conditional sub-step update to this PR.

wolenetz: I reworded some of the existing steps to make them more step-like and retain this note about 'ended' as a non-normative note since the steps are now normatively clear about this. See https://github.com/w3c/media-source/pull/151/commits/1e0c2763edd0841dc06ce474ca53aeef0f04ff23, below.

wolenetz:
@jdsmith3000 Please take another look. I significantly cleaned up the text within the "Seeking" section to keep what needs to be normative, normative :)

Helpful diff: http://services.w3.org/htmldiff?doc1=http%3A%2F%2Frawgit.com%2Fw3c%2Fmedia-source%2Fdc8a9a8400b2f822464482361db3d1ae84f3b1f0%2Findex.html&doc2=http%3A%2F%2Frawgit.com%2Fw3c%2Fmedia-source%2F1e0c2763edd0841dc06ce474ca53aeef0f04ff23%2Findex.html

index.html
```diff
@@ -1795,34 +1795,42 @@ <h4 id="h-mediasource-detach" resource="#h-mediasource-detach"><span property="x
...
             <div class="note">
-              <div class="note-title marker" aria-level="5" role="heading" id="h-note22"><span>Note</span></div>
-              <p class="">If the <code><a href="#dom-readystate">readyState</a></code> attribute is <code><a href="#idl-def-ReadyState.ended">"ended"</a></code>, then the last <code><a href="https://www.w3.org/TR/html51/semantics-embedded-content.html#timeranges-timeranges">TimeRange</a></code> in a non-empty <code><a href="https://www.w3.org/TR/html51/semantics-embedded-content.html#dom-htmlmediaelement-buffered">HTMLMediaElement.buffered</a></code> may have an end time beyond coded frame presentation intervals in one or more of the currently selected or enabled <a href="#track-buffer">track buffers</a>. Seeks must still complete to all times within that range.</p>
+              <div class="note-title marker" aria-level="5" role="heading" id="h-note23"><span>Note</span></div>
+              <p class="">In this case, one or more of the objects in <code><a href="#dom-mediasource-activesourcebuffers">activeSourceBuffers</a></code> is missing <a href="#media-segment">media segments</a> for the <var>new playback position</var>.</p>
```
jdsmith3000: I believe this note isn't particularly needed.  The first note above is useful, and your summary note below the Continue step seems useful as well.

jdsmith3000:
The section is pretty note heavy as modified here; but I think the first and last notes are good.  The one below doesn't seem necessary to me:

> In this case, one or more of the objects in activeSourceBuffers is missing media segments for the new playback position.

Otherwise, LGTM.

index.html
wolenetz: Ok. I'll fix and merge this shortly.

media-source-respec.html
```diff
@@ -705,10 +705,9 @@
           <h4>Seeking</h4>
           <p>Run the following steps as part of the "<i>Wait until the user agent has established whether or not the media data for the new playback position is available, and, if it is, until it has decoded enough data to play back that position"</i> step of the <a def-id="hme-seek-algorithm"></a>:</p>
           <ol>
-            <li>The media element looks for <a def-id="media-segments"></a> containing the <var>new playback position</var> in each <a>SourceBuffer</a> object in <a def-id="activeSourceBuffers"></a>.
+            <p class="note">The media element looks for <a def-id="media-segments"></a> containing the <var>new playback position</var> in each <a>SourceBuffer</a> object in <a def-id="activeSourceBuffers"></a>. Any position within a <a def-id="timerange"></a> in the current value of the <a def-id="hme-buffered"></a> attribute has a buffered media segment for that position.</p>
```
wolenetz: I will also fix a typo introduced here:
"has a buffered media segment" would be better as "has all necessary media segments buffered"

wolenetz:
Thanks for the review. I'll merge this shortly.

Issue #74, opened 2016-05-03
MSE Registry: Establish a update process & home


plehegar:
We have several documents for the MSE registry:
http://w3c.github.io/media-source/byte-stream-format-registry.html
http://w3c.github.io/media-source/webm-byte-stream-format.html
http://w3c.github.io/media-source/isobmff-byte-stream-format.html
http://w3c.github.io/media-source/mp2t-byte-stream-format.html
http://w3c.github.io/media-source/mpeg-audio-byte-stream-format.html

Following https://github.com/w3c/encrypted-media/issues/104 , we should publish as Notes at

https://www.w3.org/TR/mse-byte-stream-format-registry
https://www.w3.org/TR/mse-webm-byte-stream
https://www.w3.org/TR/mse-isobmff-byte-stream
https://www.w3.org/TR/mse-mp2t-byte-stream
https://www.w3.org/TR/mse-mpeg-audio-byte-stream

plehegar:
Once I managed to get the Notes published, I'll go around and ensure we have proper redirects in the right places.

paulbrucecotton:
@plehegar - I have issued an [HME WG CfC ](https://lists.w3.org/Archives/Public/public-html-media/2016May/0035.html) on this matter.  The CfC closes on Tue May 24. Please go ahead with your preparatory work and schedule the publications at a convenient time after May 24.

paulbrucecotton:
[WG Decision](https://lists.w3.org/Archives/Public/public-html-media/2016May/0057.html) supports the publication of the MSE Registry as WG Notes.

wolenetz:
Closing this v1 issue. If we need to revisit in VNext or later milestone, we can have a new tracking issue.

PR #154, opened 2016-08-18
Fix #124: Plan B: Disallow duration change to middle of any buffered coded frame

### LINKED ISSUE #124: Confusing duration change consequences and inconsistent outcome for user agents.

wolenetz:
To accomplish plan B (from discussion in #124), also includes:
* Note in duration attribute setter that the desired duration might be
  automatically increased.
* Updates to both the duration change and the end of stream algorithms'
  calculations of "highest end time" to disallow partial frame rendering
  even in muxed SourceBuffers, by using highest track buffer range end
  time, not highest buffered end time (since the latter is the result of
  track buffer range intersections and might result in a lower end time
  that truncates track buffer ranges).

### DIFF
--- a/index.html
+++ b/index.html
@@ -1372,6 +1372,10 @@ interface <span class="idlInterfaceID">MediaSource</span> : <span class="idlSupe
             <li>Run the <a href="#duration-change-algorithm">duration change algorithm</a> with <var>new duration</var> set to the value being assigned to this attribute.
               <div class="note">
                 <div class="note-title marker" aria-level="4" role="heading" id="h-note8"><span>Note</span></div>
+                <p class="">The <a href="#duration-change-algorithm">duration change algorithm</a> will adjust <var>new duration</var> higher if there is any currently buffered coded frame with a higher end time.</p>
+              </div>
+              <div class="note">
+                <div class="note-title marker" aria-level="4" role="heading" id="h-note9"><span>Note</span></div>
                 <p class=""><code><a href="#dom-sourcebuffer-appendbuffer">appendBuffer()</a></code> and <code><a href="#dom-mediasource-endofstream">endOfStream()</a></code> can update the duration under certain circumstances.</p>
               </div>
             </li>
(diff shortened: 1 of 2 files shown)


wolenetz:
@jdsmith3000 Please take a look.
@jyavenard: FYI - please raise any concerns you may have. Note especially that I refined the existing spec logic in each of "duration change algorithm" and "end of stream algorithm" to use the highest track buffer ranges end time, not the highest SourceBuffer.buffered end time. See the commit message, above, for reasoning.

@paulbrucecotton: FYI - I think we'll probably achieve consensus shortly on this landing this PR, which should fix the last currently-known V1-related spec issue (outside of @plehegar doing the non-blocking #74 fix).

jyavenard:
@wolenetz SGTM. Firefox will have to make some adjustments, as right now it was setting the duration to the highest sourcebuffer.buffered end time as per spec.

The new behaviour makes more sense too.
We will update our code accordingly.

great work.

jdsmith3000:
There's a nuance to the new language that I don't quite understand.  When buffered is read, it is set to an intersection range of all the track buffer ranges for all track buffers managed by the current SourceBuffer.  Your current change, rather than set duration to the highest endpoint of this buffered timerange, instead sets it to the highest endtime of 

> the largest track buffer ranges end time across all the track buffers across all SourceBuffer objects in sourceBuffers.

How is this different?  I don't object to the change in general, but the previous buffered language seemed equivalent and reasonably intuitive.

wolenetz:
@jdsmith3000 (https://github.com/w3c/media-source/pull/154#issuecomment-241059943) 
Intersection of track buffer ranges can result in an intersection highest end time that is less than the largest track buffer range end time. The point of using the largest track buffer range end time (instead of the largest HTMLMediaElement.buffered end time) is to *not* allow duration to ever be less than the end time of *any* currently buffered coded frame.

jdsmith3000:
@Wolenetz  I get it now.  I wasn't thinking the language through accurately.

LGTM.

wolenetz:
Thanks for reviews. And thank you, @jyavenard, for making sure #124 got filed in time to get fixed for V1 :)
I'll merge this shortly.

jyavenard:
@jdsmith3000 wrote "How is this different? I don't object to the change in general, but the previous buffered language seemed equivalent and reasonably intuitive."

I think the important part is that the duration will be set to same end value the buffered attribute would be if readyState was "ended".
With the new wording, duration will never ever be less than the latest end time of any samples in any of the tracks. The way it should be.

PR #158, opened 2016-09-09
Fix #157: Remove 'audio/mp2t' from byte stream format registry

### LINKED ISSUE #157: MPEG-2 TS Byte Stream Format MIME mismatches those in the format registry
In https://w3c.github.io/media-source/mp2t-byte-stream-format.html: "video/MP2T" is the only MIME type/subtype allowed,

but in https://w3c.github.io/media-source/byte-stream-format-registry.html, both of the following refer to that bytestream spec:
audio/mp2t
video/mp2t

So a couple questions:
1. Is the subtype case-sensitive (e.g., only allow "MP2T", or also allow "mp2t"?)
2. Is **audio/mp2t** allowed or not?

Chrome (on subset of platforms for which Chrome contains support for mp2t):
1. is case-insensitive, and
2. does not allow audio/mp2t. Just video/mp2t.

I don't believe this blocks MSE v1, but would be good to get clarified soon. I think we can make this clarification out-of-band (since the registries aren't really tied directly to v1 or VNext), but I'm marking V1NonBlocking to get appropriate attention ASAP.

wolenetz:
Following the example of the main MSE spec, this commit also includes
linking to the current w3c-respec-common.js in the format registry
respec source because saving the snapshot was otherwise failing with the
w3c-respec.common.js in the repo.

### DIFF
--- a/byte-stream-format-registry-respec.html
+++ b/byte-stream-format-registry-respec.html
@@ -3,7 +3,8 @@
   <head>
     <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
     <title>Media Source Extensions Byte Stream Format Registry</title>
-    <script src="respec-w3c-common.js" class="remove"></script>
+    <script src="https://www.w3.org/Tools/respec/respec-w3c-common" class="remove"></script>
+    <!--<script src="respec-w3c-common.js" class="remove"></script>-->
     <script src="media-source.js" class="remove"></script>
     <script class="remove">
       var respecConfig = {
@@ -182,7 +183,6 @@
           </tr>
           <tr>
             <td>
-              audio/mp2t<br>
               video/mp2t
             </td>
             <td><a href="mp2t-byte-stream-format.html">MPEG-2 Transport Streams Byte Stream Format</a></td>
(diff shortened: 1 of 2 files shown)


wolenetz:
This is one option to fix #157 

@jdsmith3000 @mwatson2 @jyavenard Please take a look. If implementations *do* currently support audio/mp2t, then maybe this isn't the right thing to do. But if they don't, this makes the format registry align with the current mp2t bytestream format spec (and the registry and spec can be adjusted later if folks at that time want to add audio/mp2t).

@plehegar @paulbrucecotton FYI

wolenetz:
Since the w3c-respec-common.js part of this change made for a large html diff, here's a helpful link for review of the generated html before and after this PR: http://services.w3.org/htmldiff?doc1=http%3A%2F%2Frawgit.com%2Fw3c%2Fmedia-source%2Feeddcad0b10a48cd633ac7962c493bec854d7035%2Fbyte-stream-format-registry.html&doc2=http%3A%2F%2Frawgit.com%2Fw3c%2Fmedia-source%2F7005cecb7585c98dd80bf6649e70bc88e3911de9%2Fbyte-stream-format-registry.html

wolenetz:
@jdsmith3000 @mwatson2 @jyavenard friendly ping

jdsmith3000:
Edge appears to return true for MediaSource .isTypeSupported() with an audio/mp2t MIIME type.  I am checking on whether it is actually supported.

jyavenard:
while Firefox doesn't support mpeg-ts ; if we are to keep video/mp2t, wouldn't it make sense to also keep audio only streams?

easier for people to transition from HLS

wolenetz:
Closing in favor of @jdsmith3000 planning to do the alternative addition of audio/mp2t to the bytestream spec to fix #157.

PR #164, opened 2016-09-16
Fix #163: Replace incorrect reference to highest-end-timestamp

### LINKED ISSUE #163: Incorrect reference to "highest end time" in a non-normative note
Current text:

> If the readyState attribute is "ended" and the new playback position is within a TimeRange currently in HTMLMediaElement.buffered, then the seek operation must continue to completion here even if one or more currently selected or enabled track buffers has a highest end timestamp less than new playback position. This condition should only occur due to logic in buffered when readyState is "ended".

"highest end timestamp" links to the track buffer definition, which is specifically used in the coded frame processing algorithm and is relevant only to the **currently in-process coded frame group**, not the entire track buffer. The quoted non-normative note, however, really means to reference the "largest range end timestamp".

I believe this is a simple change and I'll produce a PR shortly to fix it.

wolenetz:


wolenetz:
@jdsmith3000 @mwatson2 Please take a look.
@plehegar @paulbrucecotton This change should purely be an editorial correction to a non-normative note. As such, it shouldn't change the in-progress PR CfC but should be included in the document sent in the eventual request for transition to PR.

paulbrucecotton:
I agree that this an editorial correction that we can safely make to the specification before we forward the candidate Proposed Recommendation to the W3C Director.

/paulc
HME WG Chair

jdsmith3000:
LGTM.  Merge away!

PR #167, opened 2016-09-26
Fix #159: Remove decode flag from calls to Append Error Algorithm

### LINKED ISSUE #159: Unecessary argument provided to Append Error Algorithm
https://w3c.github.io/media-source/index.html#sourcebuffer-append-error

Set 5 states:
"If decode error is true, then run the end of stream algorithm with the error parameter set to "decode"."

the Append Error Algorithm is always called with the decode error parameter set to true.

Step 5 should be removed and replace the nine instances stating "run the append error algorithm with the decode error parameter set to true" (every instance referring to the append error algorithm) with "run the append error algorithm"

jdsmith3000:


jdsmith3000:
Per guidance from @paulbrucecotton, I'm staging this PR now, but it should not be merged until after CfC closes but BEFORE we send in the transition request.

media-source.js
```diff
@@ -130,9 +130,6 @@
 
   function append_error_helper(doc, df, id, text) {
     link_helper(doc, df, id, 'append error algorithm');
```
wolenetz: With this helper changed, I'd expect to see more changes in the generated html.

jdsmith3000:
I hope this is rebased now.  Please check.  I'm out this afternoon unfortunately, so if you can nudge this to the finish, please do.

wolenetz:
Looking now.

media-source.js
wolenetz: updated patch lgtm

wolenetz:
LGTM. I'll merge this now.

PR #169, opened 2016-10-03
Various changes for the publication of registries #74

plehegar:
This aligns with how we're doing with the EME registries

### DIFF
--- a/byte-stream-format-registry-respec.html
+++ b/byte-stream-format-registry-respec.html
@@ -3,7 +3,8 @@
   <head>
     <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
     <title>Media Source Extensions Byte Stream Format Registry</title>
-    <script src="respec-w3c-common.js" class="remove"></script>
+    <script src="https://www.w3.org/Tools/respec/respec-w3c-common" class="remove"></script>
+    <!--<script src="respec-w3c-common.js" class="remove"></script>-->
     <script src="media-source.js" class="remove"></script>
     <script class="remove">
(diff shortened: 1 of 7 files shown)


byte-stream-format-registry-respec.html
```diff
@@ -169,31 +170,31 @@
               audio/webm<br>
               video/webm
             </td>
-            <td><a href="webm-byte-stream-format.html">WebM Byte Stream Format</a></td>
+            <td><a def-id="byte-stream-format-registry-webm"></a> [[!MSE-FORMAT-WEBM]]</td>
```
wolenetz: This means a normative reference. I think that's correct, since the MSE spec takes only an informative reference on this registry.

media-source.js
```diff
@@ -422,10 +432,32 @@
...
+
     for (var x in groupBaseURLs) {
-      if (groupBaseURLs[x] == MSE_spec_url && window.respecConfig.specStatus == "ED") {
+      if (groupBaseURLs[x] == original_MSE_spec_url && specStatus == "ED") {
         MSE_spec_url = "index.html";
         groupBaseURLs[x] = MSE_spec_url;
+
+        // Refer to the local file rather than the published path.
+        var file = "index.html";
+        EME_spec_url = file;
```
wolenetz: nit: s/EME/MSE/

media-source.js
```diff
@@ -422,10 +432,32 @@
...
+        groupBaseURLs[x] = EME_spec_url;
```
wolenetz: nit ditto

PR #213, opened 2018-05-22
Fix #180-timestamp: Regenerated webm bytestream html to update date

### LINKED ISSUE #180: Add new vp09 codec string to WebM Byte Stream Format

wolenetz:
PR#182 contained a date from last year, though it finally landed today.
This change updates that date using a regenerated-from-respec html.

### DIFF
--- a/webm-byte-stream-format.html
+++ b/webm-byte-stream-format.html
@@ -1,429 +1,325 @@
-<!DOCTYPE html>
-<html lang="en" dir="ltr" typeof="bibo:Document " prefix="bibo: http://purl.org/ontology/bibo/ w3p: http://www.w3.org/2001/02pd/rec54#">
-<head>
-  <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
-  <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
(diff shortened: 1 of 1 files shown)


wolenetz:
@plehegar FYI HTML diff shows the changes (Editor's draft date, copyright date and text, references to governing and operating policies in SOTD). I'll land this; if there's an issue, we can correct prior to publishing as an updated note.

PR #234, opened 2019-08-20
Recommend BlockGroup over SimpleBlock in WebM bytestream spec

wolenetz:
Problems with determining coded frame duration of SimpleBlock frames in WebM have not been alleviated by WebM muxer guidelines (e.g. to use BlockGroups with BlockDurations for the last block in each track in each cluster, to avoid having to guess the duration). Furthermore, low-latency buffering+decoding of webm motivates an additional note to use BlockGroups with explicit BlockDurations instead of SimpleGroups for all blocks in a cluster (so as to not have to wait until the *next* block for each track to compute the timestamp delta and use it as the duration).

This change adds a couple non-normative notes to the current editors draft of the MSE WebM byte stream format spec.  Updating these bytestream specs can be done out-of-band relative to the main normative MSE spec, since these bytestream formats are non-normatively referenced from the bytestream format registry (and per previous repeated confirmations with W3C).

A helpful w3c-diff of this change versus the current editor's draft is available via https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fmedia-source%2Fwebm-byte-stream-format.html&doc2=https%3A%2F%2Frawgit.com%2Fwolenetz%2Fmedia-source%2Fnote_simpleblock_webm%2Fwebm-byte-stream-format.html (while my branch is alive until after this PR lands in main w3c repo).

@plehegar  Please confirm my understanding. Also, the WebM bytestream spec's editors draft was already updated since the most recent TR publishing of it. This will further update it. What is the process for getting the published (not just editor's draft) versions of these various bytestream format specs updated?

@jyavenard, @johnsim, @mwatson2, @jernoble, @chcunningham - I would appreciate your review of this pull request.

### DIFF
--- a/webm-byte-stream-format-respec.html
+++ b/webm-byte-stream-format-respec.html
@@ -251,6 +251,24 @@
           of an <a def-id="webm-init-segment"></a> is encountered.</li>
       </ol>
 
+      <div class="note">
+        Implementations MAY guess the coded frame presentation duration for
+        SimpleBlocks at the end of a <a def-id="webm-cluster"></a>.
+        Content providers are recommended to use BlockGroups with explicit
+        BlockDurations, instead of SimpleBlocks, for the last block for each
+        track in a <a def-id="webm-cluster"></a> to avoid variance in
+        buffering behavior across implementations.
+      </div>
+
+      <div class="note">
+        Implementations MAY introduce buffering latency for each SimpleBlock
+        encountered in a <a def-id="webm-cluster"></a>, since such a block's
+        coded frame presentation duration may not be determinable until the
+        next block is parsed. To avoid such latency, content providers are
+        recommended to use BlockGroups with explicit BlockDuration instead
+        of SimpleBlocks within a <a def-id="webm-cluster"></a>.
+      </div>
+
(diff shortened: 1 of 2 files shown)


tidoust:
> @plehegar Please confirm my understanding. Also, the WebM bytestream spec's editors draft was already updated since the most recent TR publishing of it. This will further update it. What is the process for getting the published (not just editor's draft) versions of these various bytestream format specs updated?

The Media WG is now responsible for the [maintenance of these specs](https://www.w3.org/2019/05/media-wg-charter.html#other-deliverables). To publish a new version, we need a group resolution. Per [charter](https://www.w3.org/2019/05/media-wg-charter.html#decisions), that means we first need to issue a call for consensus.

(Note that we will also need to update the boilerplate in the Status of This Document section to point to the Media WG instead of the HTML Media Extensions WG)

tidoust:
The WebM Byte Stream Format spec has moved to a dedicated repo. I re-created that PR there, see https://github.com/w3c/mse-byte-stream-format-webm/pull/1

Closing this PR without merging accordingly.

PR #248, opened 2020-01-24
Move former editors to their own section and update Jerry Smith status.

mounirlamouri:

PR #249, opened 2020-01-24
Separate the former and current editors

wolenetz:
The updated snapshot (index.html) includes various new respec issues
since it has not been updated in quite a while. Fixing those issues will
occur in later commits.


wolenetz:
This is a more durable alternative than https://github.com/w3c/media-source/pull/248 (this updates the spec source too, not just the generated spec).

wolenetz:
If there is no comment nor objection to this PR, I'll merge it in a week.

wolenetz:
Abandoning this PR. https://github.com/w3c/media-source/pull/248 is being updated to be more durable like this one.

PR #252, opened 2020-01-30
Add support for PR-Previews for the main MSE spec

wolenetz:
This only adds previews and diffs for the main MSE spec src file (media-source-respec.html), not the various bytestream format registry and format specs. See https://github.com/tobie/pr-preview/issues/18 for potentially adding support for previewing PRs for multiple specs within a repo.

### DIFF
--- a/.pr-preview.json
+++ b/.pr-preview.json
@@ -1,0 +1,4 @@
+{
+    "src_file": "media-source-respec.html",
+    "type": "respec"
+}


tidoust:
Byte stream format specs have now moved to dedicated repositories (with PR preview enabled) so we no longer need to worry about finding a way to enable PR preview for multiple specs in this repo. Merging this as-is to enable PR preview on the MSE spec.

wolenetz:
@tidoust thank you

PR #257, opened 2020-09-23
Fixes #197 - add application/mp4 to registry and isobmff bytestream format

### LINKED ISSUE #197: Add application/mp4 mimetype to ISOBMFF bytestream specification
The [ISOBMFF bytestream specification](https://www.w3.org/TR/mse-byte-stream-format-isobmff/) references `audio/mp4` and `video/mp4` mime types, but not `application/mp4` which is required for timed text support (and any other non audio/video content according to [RFC4337](https://tools.ietf.org/html/rfc4337#section-3.3)).

I believe that subtitles encapsulated as described by ISO/IEC 14496-30, for example, already meet all the requirement in the bytestream specification, so it would be great if `application/mp4` could be called out explicitly in  [MIME Parameters](https://www.w3.org/TR/mse-byte-stream-format-isobmff/#mime-parameters) section in the bytestream specification, and added to the mapping in the ISOBMFF entry of the [Registry](https://www.w3.org/TR/mse-byte-stream-format-registry/#registry) section of the Format Registry.

CMAF and DASH-IF, for example, call out this as the relevant mime type for timed text, so I believe this change should be in scope.

Am I right in thinking that changes can be made to the registry and bytestream specifications outside of charter as long as there is consensus?

davemevans:
Changes as proposed in #197 

I only changed the `-respec.html` files as there didn't seem to be a documented build process. I wondered if it was the same as the old encrypted-media process, but the diffs were huge.

I left the existing ReSpec and syntax errors for the editors to consider.

### DIFF
--- a/byte-stream-format-registry-respec.html
+++ b/byte-stream-format-registry-respec.html
@@ -176,7 +176,8 @@
           <tr>
             <td>
               audio/mp4<br>
-              video/mp4
+              video/mp4<br>
+              application/mp4
             </td>
             <td><a def-id="byte-stream-format-registry-isobmff"></a> [[!MSE-FORMAT-ISOBMFF]]</td>
             <td>false</td>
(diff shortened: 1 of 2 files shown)


tidoust:
Registry and Byte Stream Format specs have been moved to dedicated repositories, I re-created this pull request in the new repositories, see https://github.com/w3c/mse-byte-stream-format-isobmff/pull/1 and https://github.com/w3c/mse-byte-stream-format-registry/pull/1

Closing this PR without merging accordingly.

Issue #239, opened 2019-09-26
Drop generated versions from the repo?


tidoust:
@wolenetz, @mwatson2,

There was some discussion during TPAC on switching MSE (and EME) to the latest version of ReSpec. There is no a priori problem doing so, although some editorial updates are likely going to be necessary. Happy to help if needed.

On top of that, I note that the structure of the repository is a bit unusual. Common practice with Respec documents is to only commit the source spec. This repository contains both the source (files ending with `-respec`) and the generated version of the spec. There may be good reasons why things got done that way. I personally find it a bit clumsy since it means both representations need to be kept in sync manually.

Do you want things to stay that way? If not, I suggest to replace the generated versions with the source files, and to get rid of the `-respec` files. Happy to prepare a PR for that, but I wanted to get your perspective first.

Also, what is the difference between `index.html` and `media-source.html`?

wolenetz:
Hmm. Up until the most recent commit to media-source.html (8b236edf0be9f8dc2c0f2d6a14b6c37332d479a3), it had simply redirected to `index.html`.
@plehegar, was this an oversight? (We historically kept the draft in index.html, such that it would be visible directly via `https://w3c.github.io/media-source/`.)

@acolwell  et al setup the original structure, and I presume there was some guidance around that at the time which may differ now. I found the generated versions useful since they were snapshots that could be seen better in historical context (for instance, as respec versions/external links/etc changed, generated versions' history would show what the full context for that version of the spec was).

Keeping the generated versions updated required some editorial overhead, but not extreme IMHO.

wolenetz:
Regarding updating to the latest version of respec, I agree it'll need to be done. This is partly why the generated spec snapshots can be helpful (they highlight respec and other issues earlier). Another example of how generated spec snapshots are helpful is they can be htmldiff'ed to highlight pertinent portions of the generated result that change across the diff.

I'm certainly open to suggestions or new workflows though. Please keep these questions and suggestions/recommendations coming!

tidoust:
I confess I don't really understand why generated snapshots can help highlight respec and other issues earlier. Having things dynamic means that new ReSpec warnings/errors may pop up at any time even when the raw source does not change. But then that also means that these warnings/errors are reported as soon as possible, instead of only when merging a PR. That's all up to you in the end, I don't mean to impose anything.

[...]

PR #258, opened 2020-09-29
Use absolute URL for MSE in xref script, and fix broken fragments

tidoust:
The `media-source.js` script is used to manage cross-references in MSE specs. To be able to reference this script from dedicated repositories for byte stream formats and registry specifications (see #239), the script must use an absolute URL for the ED of the main Media Source Extensions spec.

The update also fixes a few broken fragments (naming rules have changed in Respec in the meantime).


tidoust:
Taking the liberty to merge to make progress on the creation of dedicated repos for byte stream formats specs.

wolenetz:
Thank you.

PR #263, opened 2020-11-24
Switch to Process 2020

tidoust:
This update makes a (somewhat) minimal set of changes to have the Status of this Document section say that the document is governed by Process 2020.

In theory, that should be as simple as switching to the right profile of ReSpec. In practice, that version of ReSpec changed a few things on the way definitions are handled, requiring a few other updates to definitions and links for ReSpec not to get linking errors and warnings from ReSpec.

One of these updates is using the right shorthand to link back occurrences of `readyState` (the attribute) and `ReadyState` (the enum). ReSpec now supports case sensitivity for IDL terms, and the workaround implemented in `media-source.js` was not perfect in any case (in the published Rec, an instance of "ReadyState" links back to the attribute definition in the IDL block instead of to the definition of the enum).

***
<a href="https://pr-preview.s3.amazonaws.com/tidoust/media-source/pull/263.html" title="Last updated on Nov 24, 2020, 7:01 PM UTC (936dc7a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/media-source/263/082c42b...tidoust:936dc7a.html" title="Last updated on Nov 24, 2020, 7:01 PM UTC (936dc7a)">Diff</a>

### DIFF
--- a/media-source-respec.html
+++ b/media-source-respec.html
@@ -3,8 +3,8 @@
 <head>
     <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
     <title>Media Source Extensions™</title>
-    <script src="https://www.w3.org/Tools/respec/respec-w3c-common" class="remove"></script>
-    <!--<script src="respec-w3c-common.js" class="remove"></script>-->
+    <script src="https://www.w3.org/Tools/respec/respec-w3c" class="remove"></script>
+    <script src="https://code.jquery.com/jquery-3.5.1.min.js" integrity="sha256-9/aliU8dGd2tb6OSsuzixeV4y/faTqgFtohetphbbj0=" crossorigin="anonymous"></script>
     <script src="media-source.js" class="remove"></script>
     <script class="remove">
       var respecConfig = {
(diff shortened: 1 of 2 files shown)


media-source-respec.html
```diff
@@ -3,8 +3,8 @@
 <head>
     <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
     <title>Media Source Extensions™</title>
-    <script src="https://www.w3.org/Tools/respec/respec-w3c-common" class="remove"></script>
-    <!--<script src="respec-w3c-common.js" class="remove"></script>-->
+    <script src="https://www.w3.org/Tools/respec/respec-w3c" class="remove"></script>
+    <script src="https://code.jquery.com/jquery-3.5.1.min.js" integrity="sha256-9/aliU8dGd2tb6OSsuzixeV4y/faTqgFtohetphbbj0=" crossorigin="anonymous"></script>
```
wolenetz, line 7: Why is this necessary? Is the deprecation note at https://respec.org/docs/#postProcess outdated such that respec no longer includes jQuery? It seems a bugfix to remove jQuery dependency in the various .js for this spec would be good, regardless. WDYT?

wolenetz (approved):
Generally LGTM.

media-source-respec.html
tidoust: Indeed, the deprecation note is now outdated, jQuery was dropped in the new profile, see comment at the beginning of https://github.com/w3c/respec/blob/develop/profiles/w3c-common.js

I reported the need to update the doc in https://github.com/w3c/respec/issues/3228

Regarding dropping jQuery, I agree. I'm proceeding in steps :)

For the time being, `media-source.js` needs jQuery. I'll see how to drop that dependency. Actually, I believe that we can aim at dropping most of `media-source.js` in the end. Cross-references are now pretty well supported in ReSpec, so we should be able to restrict the need to maintain custom code to a bare minimum. Also, relying on ReSpec would allow us to detect problems (such as definitions that other specs don't export) more easily, and to benefit from features such as the context menu that lists references to a given term.

wolenetz: All SGTM. Thank you.

PR #265, opened 2020-11-29
Convert most references to ReSpec's shortlinks syntax

tidoust:
This pull request converts most of the occurrences of `def-id` in the spec to ReSpec's shorthands syntax. Motivation behind the changes:
- Makes the source slightly more readable and easier to edit (wrapping characters are still needed but no need to deal with IDs anymore or no need to add custom code).
- Less maintenance on custom script.
- The "right" version of external specs gets used, e.g the HTML LS instead of HTML5.1, or the ED of WebIDL instead of the published version.
- ReSpec computes references to a dfn, showed in a popup window when one clicks on a definition.
- It will be easier to detect terms from external specs that no longer exist or that these specs don't "export".

The pull request is rather large, but a commit approach should ease the review:
1. https://github.com/w3c/media-source/commit/14fc4aeda188a217dfa0a8bbc2f6c10a3f4bd8cb makes manual updates to the source to prepare for the conversion.
2. https://github.com/w3c/media-source/commit/281fe9c05ca36e2537a8934d6aa07d9c21b472ed applies a [conversion script](https://gist.github.com/tidoust/f8cef476ef8f8e43af1b5ec04f970747) to the prepared source to convert all `def-id` occurrences that can be easily processed automatically.
3. https://github.com/w3c/media-source/commit/525d2e5fb50d104987323666f67ffcf9dce16320 drops the entries from `media-source.js` that are no longer needed after the conversion, leaving only those that still need to be converted AND those that byte stream format specs still reference so that these specs still render properly while they get updated.

The conversion leaves about 150 occurrences of `def-id` (out of the 1200 initial ones) that will still need to be addressed over time. They correspond to terms in external specs that are not yet part of the xref database used by ReSpec for some reason (terms in File API, some enum values and internal terms in the HTML spec), or to terms that need some custom logic.

The whole thing is purely editorial.

***
<a href="https://pr-preview.s3.amazonaws.com/tidoust/media-source/pull/265.html" title="Last updated on Apr 28, 2021, 1:06 PM UTC (32be759)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/media-source/265/6832c8b...tidoust:32be759.html" title="Last updated on Apr 28, 2021, 1:06 PM UTC (32be759)">Diff</a>


saschanaz:
LGTM to me, @wolenetz could you review this while you are at it? (I mean, you just found #270 😀)

wolenetz:
I expect to review this one tomorrow. Thank you for doing this, @tidoust !

tidoust:
I rebased the pull request against the main branch and forced pushed an update.

(There's a hiccup in the first commit where the end of the document gets duplicated for some reason. That's fixed in the second commit)

wolenetz:
@ https://github.com/w3c/media-source/pull/265#event-4656732602 - thank you, I'll get back to work on this review today. I've been out of office.

media-source-respec.html
```diff
@@ -2332,6 +2332,69 @@ <h2>Examples</h2>
 
 &lt;video id="v" autoplay&gt; &lt;/video&gt;
 
+&lt;script&gt;
```
wolenetz: This appears to begin the duplicative hiccup noted in the conversation (https://github.com/w3c/media-source/pull/265#issuecomment-828443753)  and should be fixed by the next commit in this sequence.

wolenetz: Yes, it looks like this hiccup is fixed by the second commit.

media-source-respec.html
```diff
@@ -229,30 +231,30 @@ <h3>Definitions</h3>
...
           </dd>
           <dt><dfn id="presentation-interval">Presentation Interval</dfn></dt>
           <dd>
-            <p>The presentation interval of a <a def-id="coded-frame"></a> is the time interval from its <a def-id="presentation-timestamp"></a> to the <a def-id="presentation-timestamp"></a> plus the <a def-id="coded-frames-duration"></a>. For example, if a coded frame has a presentation timestamp of 10 seconds and a <a def-id="coded-frame-duration"></a> of 100 milliseconds, then the presentation interval would be [10-10.1). Note that the start of the range is inclusive, but the end of the range is exclusive.</p>
+            <p>The presentation interval of a [=coded frame=] is the time interval from its [=presentation timestamp=] to the [=presentation timestamp=] plus the [=coded frame duration|coded frame's duration=]. For example, if a coded frame has a presentation timestamp of 10 seconds and a [=coded frame duration=] of 100 milliseconds, then the presentation interval would be [10-10.1). Note that the start of the range is inclusive, but the end of the range is exclusive.</p>
```
wolenetz, line 238: How strong is the disrecommendation for concept aliasing in respec? Is a rephrase to eliminate the need for this alias here a good idea (if so, I cann do in a later commit).

wolenetz (commented):
Looks good to me. Thank you for modernizing the shorthand for so many of these links in the spec!

media-source-respec.html
tidoust: Aliasing is [not recommended](https://github.com/w3c/respec/wiki/Shorthands-Guide#aliasing-concepts). Whether you should view that disrecommendation as strong is up to you. It makes sense to alias a concept when additional contextualization is needed. In this particular case, aliasing does not seem required so getting rid of it seems like a good idea. There are 11 occurrences of "coded frame duration" and 2 occurrences of "coded frame's duration".

PR #272, opened 2021-05-12
Remove frozen files from main branch

tidoust:
This removes files that no longer need to be updated from the main branch:
- registry files that moved to their own repositories and that only contain a redirect.
- copies of published working draft, last call, candidate recommendations, proposed recommendation, that will never change
- the copy of Respec's JS which is no longer used anywhere

The goal is to end up in a situation where a quick glance at the repo on GitHub is enough to understand what it contains and where the source of the Media Source Extensions spec is.

Note all of these files will continue to exist and be served from the `gh-pages` branch. That is good. There may still exist links to them from other places on the Web.


wolenetz (approved):

wolenetz:
Thank you. I agree it is good to still have many of these on the `gh-pages` branch. For example, the REC spec's link to the bytestream format registry is ok, but if we also removed registry and formats themselves from the `gh-pages` branch, then that registry doc's links to current ED of itself would be broken (and likewise for the formats).

Issue #175, opened 2016-12-15
MediaSource in Worker


DanielBaulig:
The Videos team at Facebook has a growing need to move calls to Media Source Extensions off of the main thread and into a Worker. While most parts of your off-the-shelve streaming video player can be implemented in a Worker (fetching and parsing of the manifest as well as selecting and fetching of segments) the calls to and interactions with the Media Source Extensions API (e.g. `SourceBuffer.prototype.appendBuffer`, `SourceBuffer.prototype.remove`, etc) remain a bottleneck and have to happen on the main browser rendering thread. 

On a sizable and complex web site large blocks of JS execution can regularly block the main thread making videos slow to start playing and causing lots of rebuffering, especially in live videos where you can only buffer ahead very little data. To ensure great web video performance in the long run, even as web apps continue to grow in size and complexity, we believe we need to allow video players to move into Workers and off of the main thread.

We would like to propose this change to the working group and involve other API users as well as implementors in the discussion.

bmaurer:
#14 also seems relevant here -- based on my reading of that bug it would help with part of this problem (once you fetch a segment you'd pass a stream to that audio to MediaSource and would bypass the main thread, or even a worker thread in terms of getting that data to where it needs to go).

But that still leaves the issue of fetching a manifest and initiating a request for a segment

jyavenard:
> the calls to and interactions with the Media Source Extensions API (e.g. SourceBuffer.prototype.appendBuffer, SourceBuffer.prototype.remove, etc) remain a bottleneck and have to happen on the main browser rendering thread.

That it's a bottleneck would be mostly a user agent implementation issue.
Those APIs are designed to be asynchronous, they should take no time at all and not block the main thread.
With current Firefox, running those two operations in a worker would make little (or no) performance difference, they are in effect running internally in a worker.

I wish issue #100 progressed further, it would certainly facilitate handling those operations.

bmaurer:
The issue here isn't the cost of video *on* the main thread. The issue is that if the main thread is busy with other work (say the javascript needed to render a page) that it can not be responsive to input from the network. This means that the video will be slower to buffer even though the client has been able to get the video over the network.

The goal is for the video loading code to be isolated on a worker thread where it can operate in real time without the risk of latency induced by work on the main thread.

zhaoz:
[...]

PR #282, opened 2021-06-30
MSE-in-Workers draft feature specification

wolenetz:
Specifies the exposure of MSE API additionally to DedicatedWorkerGlobalScope.
Updates, when MSE is in such a worker scope, cross-context communication between a Window HTMLMediaElement and a DedicatedWorker MSE instance to make it clear to implementors what is necessary for interop on this feature.
Spec feature for MSE-in-Workers is #175 

Note that Chromium already has an experimental implementation and there is a demo of the feature with instructions available at https://wolenetz.github.io/mse-in-workers-demo/mse-in-workers-demo.html

***
<a href="https://pr-preview.s3.amazonaws.com/wolenetz/media-source/pull/282.html" title="Last updated on Jul 26, 2021, 10:35 PM UTC (077e233)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/media-source/282/801c698...wolenetz:077e233.html" title="Last updated on Jul 26, 2021, 10:35 PM UTC (077e233)">Diff</a>

### DIFF
--- a/media-source-respec.html
+++ b/media-source-respec.html
@@ -14,7 +14,10 @@
       // the specification's short name, as in https://www.w3.org/TR/short-name/
       shortName: "media-source",
 
-      // if there a publicly available Editor's Draft, this is the link
+      // If there's a publicly available Editor's Draft, this is the link.
+      // Though the 'github' respecConfig key now duplicates this functionality, this key
+      // is still needed for media-source.js preprocessing for detecting if the
+      // including respec source is a main spec or one of the byte-stream specs.
       edDraftURI:           "https://w3c.github.io/media-source/",
 
       // if this is a LCWD, uncomment and set the end of its review period
@@ -78,19 +81,9 @@
 
       scheme: "https",
 
+      github: "w3c/media-source",
+
       otherLinks: [{
-      key: 'Repository',
-      data: [{
-        value: 'We are on GitHub',
-        href: 'https://github.com/w3c/media-source/'
-      }, {
-        value: 'File a bug',
-        href: 'https://github.com/w3c/media-source/issues'
-      }, {
-        value: 'Commit history',
-        href: 'https://github.com/w3c/media-source/commits/gh-pages/media-source-respec.html'
-      }]
-    },{
       key: 'Mailing list',
       data: [{
         value: 'public-media-wg@w3.org',
@@ -291,13 +284,19 @@
     "closed",
     "open",
     "ended"
-};</pre><table class="simple" data-dfn-for="ReadyState"><tbody><tr><th colspan="2">Enumeration description</th></tr><tr><td><dfn><code id="idl-def-ReadyState.closed">closed</code></dfn></td><td>
+};</pre><table class="simple" data-dfn-for="ReadyState"><tbody><tr><th colspan="2">Enumeration description</th></tr>
+        <tr><td><dfn><code id="idl-def-ReadyState.closed">closed</code></dfn></td><td>
           Indicates the source is not currently attached to a media element.
-        </td></tr><tr><td><dfn><code id="idl-def-ReadyState.open">open</code></dfn></td><td>
+        </td></tr>
+        <tr><td><dfn><code id="idl-def-ReadyState.open">open</code></dfn></td><td>
           The source has been opened by a media element and is ready for data to be appended to the <a>SourceBuffer</a> objects in {{MediaSource/sourceBuffers}}.
         </td></tr><tr><td><dfn><code id="idl-def-ReadyState.ended">ended</code></dfn></td><td>
           The source is still attached to a media element, but {{MediaSource/endOfStream()}} has been called.
         </td></tr></tbody></table></div>
+
+      <div class="issue" data-number="276">
+          <p>Consider adding a "<code>closing</code>" {{MediaSource/ReadyState}} to indicate the source is in the process of being concurrently detached from a media element. This would be useful for some implementations of {{MediaSource}} and {{SourceBuffer}} in {{DedicatedWorkerGlobalScope}}.</p>
+      </div>
 
       <div><pre class="idl">enum EndOfStreamError {
     "network",
@@ -310,7 +309,7 @@
           <p class="note">JavaScript applications SHOULD use this status code to terminate playback with a decode error. For example, if a parsing error occurs while processing out-of-band media data.</p>
         </td></tr></tbody></table></div>
 
-<pre class="idl">[Exposed=Window]
+<pre class="idl">[Exposed=(Window,DedicatedWorker)]
 interface MediaSource : EventTarget {
     constructor();
     readonly        attribute SourceBufferList    sourceBuffers;
@@ -320,6 +319,7 @@ interface MediaSource : EventTarget {
                     attribute EventHandler        onsourceopen;
                     attribute EventHandler        onsourceended;
                     attribute EventHandler        onsourceclose;
+    static readonly attribute boolean canConstructInDedicatedWorker;
     SourceBuffer   addSourceBuffer (DOMString type);
     undefined           removeSourceBuffer (SourceBuffer sourceBuffer);
     undefined           endOfStream (optional EndOfStreamError error);
@@ -358,8 +358,8 @@ interface MediaSource : EventTarget {
             <li>If the value being set is negative or NaN then throw a {{TypeError}} exception and abort these steps.</li>
             <li>If the {{MediaSource/readyState}} attribute is not {{ReadyState/""open""}} then throw an {{InvalidStateError}} exception and abort these steps.</li>
             <li>If the {{SourceBuffer/updating}} attribute equals true on any <a>SourceBuffer</a> in {{MediaSource/sourceBuffers}}, then throw an {{InvalidStateError}} exception and abort these steps.</li>
-            <li>Run the [=duration change=] algorithm with <var>new duration</var> set to the value being assigned to this attribute.
-              <p class="note">The [=duration change=] algorithm will adjust <var>new duration</var> higher if there is any currently buffered coded frame with a higher end time.</p>
+            <li>Run the [=duration change=] algorithm with |new duration| set to the value being assigned to this attribute.
+              <p class="note">The [=duration change=] algorithm will adjust |new duration| higher if there is any currently buffered coded frame with a higher end time.</p>
               <p class="note">{{SourceBuffer/appendBuffer()}} and {{MediaSource/endOfStream()}} can update the duration under certain circumstances.</p>
             </li>
           </ol>
@@ -369,6 +369,13 @@ interface MediaSource : EventTarget {
           <p>The event handler for the {{sourceended}} event.</p>
         </dd><dt><dfn><code>onsourceclose</code></dfn> of type {{EventHandler}}</dt><dd>
           <p>The event handler for the {{sourceclose}} event.</p>
+        </dd>
+        <dt><dfn><code>canConstructInDedicatedWorker</code></dfn> of type {{boolean}}</dt><dd>
+          <p>Returns true.</p>
+          <p class="note">This attribute enables main thread and dedicated worker feature detection of support for
+            creating and using a {{MediaSource}} object in a dedicated worker, and mitigates the need for higher latency
+            detection polyfills like attempting creation of a {{MediaSource}} object from a dedicated worker, especially
+            if the feature is not supported.</p>
         </dd></dl></section><section><h2>Methods</h2><dl class="methods" data-dfn-for="MediaSource"><dt><dfn><code>addSourceBuffer</code></dfn></dt><dd>
           <p>Adds a new <a>SourceBuffer</a> to {{MediaSource/sourceBuffers}}.</p>
           <ol class="method-algorithm">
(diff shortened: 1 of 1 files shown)


wolenetz:
It looks like I need to fix a few spec-prod errs (div in dl, dl in ol, div in ol, div in ol, ol in dl) per https://github.com/w3c/media-source/runs/2947876193?check_suite_focus=true
Please don't let those block review.

media-source-respec.html
```diff
@@ -418,73 +422,158 @@ <h3>Attributes</h3>
               </ol>
             </li>
 
-            <li>Let <var>SourceBuffer audioTracks list</var> equal the {{AudioTrackList}} object returned by <var>sourceBuffer</var>.{{SourceBuffer/audioTracks}}.</li>
-            <li>If the <var>SourceBuffer audioTracks list</var> is not empty, then run the following steps:
+            <li>Let |SourceBuffer audioTracks list| equal the {{AudioTrackList}} object returned by
```
tidoust, line 425: It is a good idea to use the `|` shorthand syntax instead of `<var>`. This makes ReSpec track and highlight variables when user hovers or clicks on them as in https://user-images.githubusercontent.com/8426945/74150174-5adf9f00-4c2f-11ea-82ee-d7c059a8f772.gif

That may exceed the scope of this PR but I note that it would be equally good to replace constructs such as `<var>sourceBuffer</var>` with `|sourceBuffer|` throughout algorithms as well.

media-source-respec.html
```diff
@@ -369,6 +369,10 @@ <h3>Attributes</h3>
           <p>The event handler for the {{sourceended}} event.</p>
         </dd><dt><dfn><code>onsourceclose</code></dfn> of type {{EventHandler}}</dt><dd>
           <p>The event handler for the {{sourceclose}} event.</p>
+        </dd>
+        <dt><dfn><code>canConstructInDedicatedWorker</code></dfn> of type {{boolean}}</dt><dd>
+          <p>Returns true.</p>
+          <p class="note">This attribute enables feature detection of support for MSE-in-Workers without requiring an application to attempt usage first from a DedicatedWorker.</p>
```
tidoust, line 375: "MSE-in-Workers" and "DedicatedWorker" seem jargon to me. Suggest to expand with regular prose as in:

```suggestion
          <p class="note">This attribute enables feature detection of support for creating and using a {{MediaSource}} object in a dedicated worker directly from the main thread without requiring an application to attempt usage first from a dedicated worker.</p>
```

media-source-respec.html
```diff
@@ -599,6 +688,40 @@ <h3>Event Summary</h3>
...
+        in the context where it was created.</p>
+
+        The rest of this section describes a model for bounding information latency for attachments of a {{Window}}
+        media element to a {{DedicatedWorkerGlobalScope}} {{MediaSource}}. While the model describes communication using
+        message passing, implementations MAY choose to communicate in potentially faster ways, such as using shared
+        memory and locks. Attachments to a {{Window}} {{MediaSource}} synchronously have the information already without
+        communicating it across contexts.
+
+        <p>A {{MediaSource}} that is constructed in a {{DedicatedWorkerGlobalScope}} has an internal, unexposed
+        <dfn id="port-to-main">port to main</dfn> nullable variable that stores a {{MessagePort}} setup during
```
tidoust, line 711: In other specs, these internal, unexposed variable are called [internal slots](https://medium.com/jspoint/what-are-internal-slots-and-internal-methods-in-javascript-f2f0f6b38de) and rendered as `[[port to main]]`. From a ReSpec perspective, dfn would be `<dfn data-dfn-for="MediaSource">[[\port to main]]</dfn> and references would use `{{MediaSource/[[port to main]]}}`

media-source-respec.html
```diff
@@ -599,6 +688,40 @@ <h3>Event Summary</h3>
...
+        attachment and nulled during detachment. A {{Window}} {{MediaSource}}'s [=port to main=] is always null.</p>
+
+        </p>An {{HTMLMediaElement}} extended by this specification and attached to a {{DedicatedWorkerGlobalScope}}
+        {{MediaSource}} similarly has an internal, unexposed <dfn id="port-to-worker">port to worker</dfn> nullable
+        variable that stores a {{MessagePort}} setup during attachment and nulled during detachment. A
+        [=port to worker=] is always null unless attached to a {{DedicatedWorkerGlobalScope}} {{MediaSource}}.</p>
+
```
tidoust, line 718: I would also add the definition of the `channel` internal slot here as well, even though it is only going to be set in objects created in dedicated workers.

media-source-respec.html
```diff
@@ -755,7 +918,16 @@ <h4><dfn>SourceBuffer Monitoring</dfn></h4>
 
         <section id="active-source-buffer-changes">
           <h4><dfn>Changes to selected/enabled track state</dfn></h4>
-          <p>During playback {{MediaSource/activeSourceBuffers}} needs to be updated if the <a def-id="videoref" name="dom-videotrack-selected">selected video track</a>, the <a def-id="videoref" name="dom-audiotrack-enabled">enabled audio track(s)</a>, or a text track <a def-id="videoref" name="dom-texttrack-mode">mode</a> changes. When one or more of these changes occur the following steps need to be followed.</p>
+          <p>During playback {{MediaSource/activeSourceBuffers}} needs to be updated if the <a def-id="videoref" name="dom-videotrack-selected">selected video track</a>, the <a def-id="videoref" name="dom-audiotrack-enabled">enabled audio track(s)</a>, or a text track <a def-id="videoref" name="dom-texttrack-mode">mode</a> changes. When one or more of these changes occur the following steps need to be followed.
+             Also, when {{MediaSource}} was constructed in {{DedicatedWorkerGlobalScope}}, then each change that occurs
```
tidoust, line 922:
```suggestion
             Also, when {{MediaSource}} was constructed in a {{DedicatedWorkerGlobalScope}}, then each change that occurs
```

tidoust (commented):
Some comments on the syntax.

The logic looks sound on the surface (a bit convoluted but it's not obvious that it can be simplified...). I haven't tried to look into details yet.

tidoust (commented):
Looking into it some more, I wonder what are the benefits of specifying such a detailed message passing logic between the object in the worker and its mirror in Window.

Wouldn't it be enough to mention the need for implementation to mirror information between Window and worker objects using whatever implementation means they want (as alluded to in Cross-context communication model section) and keep algorithms somewhat unchanged except perhaps to note which information needs to have been mirrored and/or where the algorithm step runs?

For instance, considering the following steps:

> If the MediaSource was constructed in a DedicatedWorkerGlobalScope : Post an internal end of stream message to port to main whose implicit handler in Window notifies the media element that it now has all of the media data. 
>
> Otherwise: Notify the media element that it now has all of the media data. 

Why not keep it to what it was before the change?

> Notify the media element that it now has all of the media data.

There is only one media element. It is clear that it is attached to the Window context. The term "notify" is vague, allowing arbitrary complex implementations. I do realize that the worker case is going to be more complex to implement but, from a spec perspective, I'm not clear why specifying the extra step that cannot be directly tested in practice is needed.

(There are likely other algorithms where it remains useful to clarify what is to happen, for instance to mention that visible VideoTrack objects attached to a media element are going to be mirrors of the ones in the worker)

media-source-respec.html
wolenetz: I agree - I used `|` in much of the PR. I'll fix portions in this PR that touch lines with `<var>`, too. There are currently over 300 uses, so fixing the remaining bulk will be a different PR.

wolenetz: cf47a8f addresses this feedback. There still remain over 200 occurrences in the spec after that.

media-source-respec.html
wolenetz: Good point. 9efdf52 addresses this comment.

wolenetz:
> Looking into it some more, I wonder what are the benefits of specifying such a detailed message passing logic between the object in the worker and its mirror in Window.
> 
> Wouldn't it be enough to mention the need for implementation to mirror information between Window and worker objects using whatever implementation means they want (as alluded to in Cross-context communication model section) and keep algorithms somewhat unchanged except perhaps to note which information needs to have been mirrored and/or where the algorithm step runs?

The notion of "mirrored" can be interpreted differently in concurrent systems like the main/worker contexts in this feature. For clarity of what information coherency/latency flexibility is allowed, I included the "Cross-context communication model" and use of it (`port to worker` / `port to main`) in specific portions of the algorithms where the communication.

> For instance, considering the following steps:
> 
> > If the MediaSource was constructed in a DedicatedWorkerGlobalScope : Post an internal end of stream message to port to main whose implicit handler in Window notifies the media element that it now has all of the media data.
> > Otherwise: Notify the media element that it now has all of the media data.
> 
> Why not keep it to what it was before the change?
> 
> > Notify the media element that it now has all of the media data.
> 
> There is only one media element. It is clear that it is attached to the Window context. The term "notify" is vague, allowing arbitrary complex implementations. I do realize that the worker case is going to be more complex to implement but, from a spec perspective, I'm not clear why specifying the extra step that cannot be directly tested in practice is needed.

This example is a good one for your case. The "notify" is indeed vague due to legacy media pipelines and demuxers introducing some information latencies. I will revert the convoluted text specifically for this example. In contrast, the buffered and seekable extensions need more clarity because they are synchronous (on main thread) and reflect current information immediately, so clarify what the information latency limits here with the model and its usage seemed to me to be not only a good idea but required for better interoperability.

> (There are likely other algorithms where it remains useful to clarify what is to happen, for instance to mention that visible VideoTrack objects attached to a media element are going to be mirrors of the ones in the worker)

Indeed, like the seekable and buffered extensions, the initialization segment received algorithm and removeSourceBuffer's management of tracks is IMHO made much clearer for interop by the detailed spec. Implementors are free to replace internal, unexposed items in the spec like 'update track state' messages with equivalent.

tl;dr: I wanted to clarify the expectations for better implementation interop.

media-source-respec.html
wolenetz: Yes I've seen the internal slot spec abstraction occurring recently in other specs. I'll update this feature draft for port-to-main/window and channel to be held in explicit internal slots. Separately, in a different PR, I'll update the existing object variables like "first initialization segment received flag" to be held in internal slots, too.

media-source-respec.html
wolenetz: I agree and will update.

media-source-respec.html
wolenetz: Thanks! Fixed by 82070b2

tidoust:
> tl;dr: I wanted to clarify the expectations for better implementation interop.

Right and you shouldn't consider my comments as blocking in any way! I didn't spot anything wrong in the pull request.

The algorithms can be iterated over to drop bits that won't be observable in any way. These cases will typically appear when test cases get added to the test suite to assess interop between implementations of MSE in workers.

media-source-respec.html
wolenetz: 98fe6fc should fix this.

media-source-respec.html
wolenetz: 98fe6fc should fix this.

wolenetz:
> The algorithms can be iterated over to drop bits that won't be observable in any way. These cases will typically appear when test cases get added to the test suite to assess interop between implementations of MSE in workers.

Sounds good to me. I've added ee8fc38 to address the one known place where there was perhaps too much verbosity added.

wolenetz:
077e233 is a squashed replacement of the previous chain of commits, to ease potential rebasing.

wolenetz:
@mwatson2 please take a look. Known related work that I'd like to get landed soon, separately, is adding privacy and security sections, and especially (in response to a high-res timer attack concern raised in a response to my intent-to-experiment request in chromium/blink) clarify that the implementation's choice of cross-context communication mechanism needs to be resilient against such attacks (such as providing variable delay akin to postMessage, as the model already describes, if optimizations are made) for information such as MediaSource duration, live-seekable-range (as reported to HTMLMediaElement from worker) and recent error status (as reported to worker MSE from window HTMLMediaElement).

@tidoust please help verify that the changes in response to your comments are correct.

Thank you,
Matt

tidoust:
> @tidoust please help verify that the changes in response to your comments are correct.

That looks good to me, thanks!

I note that the IPR bot complains because Google (and other companies) needs to re-join the working group following re-chartering.

mwatson2 (approved):

mwatson2:
@wolenetz Looks good to me. My only comments are cosmetic suggestions - not technical - and could be done later, if done at all.

wolenetz:
@ https://github.com/w3c/media-source/pull/282#issuecomment-89805456 (@mwatson2) - Thank you for your review. In general, I agree with your comments. The "potentially faster" text was included to ensure that *at least* no more than the latency of a postMessage() done by the app would be the limit. Otherwise, apps would need to take yet more care to understand the state of the attachment cross-context. In this regard, I'm also currently undoing some of the optimizations taken in the Chromium prototype of this feature, as they may enable Spectre-like exploits.

Since my AC rep hasn't yet rejoined the WG, I'll await until after my vacation (all of next week) to address comments (or land as-is and do a follow-up to address comments).

wolenetz:
@https://github.com/w3c/media-source/pull/282#issuecomment-898054566 in the interests of more easily unblocking the follow-on commits that I'd been working on while pending rejoin to the WG, I'll land this as is and follow-up in later PR on your comments, above.

media-source-respec.html
```diff
@@ -404,108 +411,202 @@ <h3>Attributes</h3>
...
                     </li>
-                    <li>Remove the {{AudioTrack}} object from the <var>SourceBuffer audioTracks list</var>.
-                    <p class="note">
-                      This should trigger {{AudioTrackList}} [[HTML]] logic to <a def-id="queue-and-fire-media-removetrack-with-track-attr-initialized-to"></a> the {{AudioTrack}} object, at the <var>SourceBuffer audioTracks list</var>.
-                      If the {{AudioTrack/enabled}} attribute on the {{AudioTrack}} object was true at the beginning of this removal step, then this should also trigger {{AudioTrackList}} [[HTML]] logic to [=queue a task=] to [=fire an event=] named <a def-id="mediatracklist-change"></a> at the <var>SourceBuffer audioTracks list</var>
-                    </p>
+
+                    <li>
+                      <dl class="switch">
+                        <dt>If the {{MediaSource}} was constructed in a {{DedicatedWorkerGlobalScope}}:</dt>
```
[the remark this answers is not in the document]
wolenetz: > IIUC, the purpose of the mirror messages is to ensure that the HTML Media Element update steps run as their own task on the main event loop, rather than the associated changes happening "magically" in the middle of some other task execution, right ?

This is typically the case, yes. Especially will be more so once I update this to likely limit any optimizations of the 'cross-context communication mechanism' for web platform security and interop.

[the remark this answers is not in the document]
wolenetz: > This is by no means an essential change, but I think this would be clearer if the "Otherwise" steps below were broken out an given a name (e.g. "_Remove an audio track from Media Element audio track list_") and then this step can be something like
> 
> > If the MediaSource was constructed in a DedicatedWorkerGlobalScope, post an internal remove track mirror message ... whose implicit handler runs the _Remove an audio track from Media Element audio track list_ algorithm on ....
> > Otherwise run the _Remove an audio track from Media Element audio track list_ algorithm on ....

I'm looking to see if there's a way to keep such a "named step" within the method algorithm steps instead of being moved to an explicit algorithm in the SourceBuffer algorithm's section. Do you have any preference on this?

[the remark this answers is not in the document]
wolenetz: For consistency with other media WG specs, and internal consistency with existing MSE spec, moving such a "named step" to be an explicit algorithm is probably best.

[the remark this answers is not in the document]
wolenetz: Mirror if necessary seems the better approach for this, as it is more general.

PR #285, opened 2021-07-31
Modernize var syntax, including type info where known

wolenetz:
Updates `<var>...</var>` to instead use modern respec `|...|`, and in
most cases, on first definition of a variable, includes type of variable
as a suffix to `...` in the definition.

Links IDL method arguments in the definition to the argument variable's
detail by also wrapping those argument names in `|...|`.

Includes a couple other minor changes: uses `minus` instead of `-` in
algorithms in a couple places.

Replaces a couple instances of "source buffer" with "{{SourceBuffer}}"
for improved linkage.

Note, the MSE-in-Workers initial draft PR (#282) contains some similar
changes to lines only relevant to that draft, and will likely need
rebasing and manual merge resolution on top of this once this lands.

***
<a href="https://pr-preview.s3.amazonaws.com/wolenetz/media-source/pull/285.html" title="Last updated on Sep 1, 2021, 12:52 AM UTC (de9943a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/media-source/285/fbfb36f...wolenetz:de9943a.html" title="Last updated on Sep 1, 2021, 12:52 AM UTC (de9943a)">Diff</a>

### DIFF
--- a/media-source-respec.html
+++ b/media-source-respec.html
@@ -601,7 +601,7 @@ interface MediaSource : EventTarget {
           <ol class="method-algorithm">
             <li>If the {{MediaSource/readyState}} attribute is not in the {{ReadyState/""open""}} state then throw an {{InvalidStateError}} exception and abort these steps.</li>
             <li>If the {{SourceBuffer/updating}} attribute equals true on any <a>SourceBuffer</a> in {{MediaSource/sourceBuffers}}, then throw an {{InvalidStateError}} exception and abort these steps.</li>
-            <li>Run the [=end of stream=] algorithm with the |error:EndOfStreamError| parameter set to |error|.</li>
+            <li>Run the [=end of stream=] algorithm with the error parameter set to |error:EndOfStreamError|.</li>
           </ol>
         <table
           class="parameters"><tbody><tr><th>Parameter</th><th>Type</th><th>Nullable</th><th>Optional</th><th>Description</th></tr><tr><td class="prmName">|error|</td><td class="prmType">{{EndOfStreamError}}</td><td class="prmNullFalse"><span role="img" aria-label="False">✘</span></td><td class="prmOptTrue"><span role="img" aria-label="True">✔</span></td><td class="prmDesc"></td></tr></tbody></table><div><em>Return type: </em>{{undefined}}</div></dd>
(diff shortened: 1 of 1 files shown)


media-source-respec.html
```diff
@@ -467,51 +467,51 @@ <h3>Attributes</h3>
...
           <p>Signals the end of the stream.</p>
 
           <ol class="method-algorithm">
             <li>If the {{MediaSource/readyState}} attribute is not in the {{ReadyState/""open""}} state then throw an {{InvalidStateError}} exception and abort these steps.</li>
             <li>If the {{SourceBuffer/updating}} attribute equals true on any <a>SourceBuffer</a> in {{MediaSource/sourceBuffers}}, then throw an {{InvalidStateError}} exception and abort these steps.</li>
-            <li>Run the [=end of stream=] algorithm with the <var>error</var> parameter set to <var>error</var>.</li>
+            <li>Run the [=end of stream=] algorithm with the |error:EndOfStreamError| parameter set to |error|.</li>
```
tidoust, line 506: I note that ReSpec gets slightly confused here as it thinks that both occurrences of `error` refer to the same variable, whereas the first `|error:EndOfStreamError|` actually refers to the name of the parameter in the end of stream algorithm, whereas the second `|error|` refers to the name of the parameter of the `endOfStream` method that is being described here.

At some point in the future, there may exist a way to flag algorithms and algorithm parameters as such (see related discussion in https://github.com/speced/spec-dfn-contract/issues/2), which would provide a way to distinguish between the two. I don't have a good workaround to suggest in the meantime other than to drop the `|` around the first occurrence (to end up with an undecorated "with the error parameter"). This does not strike me as very confusing in any case, so leaving things as-is seems good too.

tidoust (approved):
Nice!

wolenetz:
Note, https://github.com/w3c/media-source/commit/3775bc8b7e6304fe9f09878c01dfa2602b44fdff was a force push of the result of manual rebase over the top of the now-landed PR#282.

wolenetz:
@mwatson2: since @tidoust has approved this and it's been out for review for a month, I'll merge it shortly. If you prefer different, we can revert and fix later.

Issue #259, opened 2020-10-20
Consider adding support for apps to get metadata about what is currently buffered


wolenetz:
A "[SourceBuffer Introspection API](https://docs.google.com/document/d/1XBSu9x-z8gHwh0R_0Py7N8OuRz9QpNAld9AvQt8NSpY/edit)" proposal was discussed at FOMS 2019 in October, but I don't currently find a spec issue tracking further discussion and potential inclusion in MSE v2, so I've filed this issue.

This API would enable greater ability to have precise interop tests, as well as give apps much more information about what is buffered so that they can workaround potential content or API implementation issues.

Comments welcome, especially: is this or something like this something which you would really want available in MSE? Are we missing your use case? Or is this feature unnecessary?

credit for the proposal: @sandersdan

wolenetz:
Including the current mimetype+codecs string verbatim from the most recent of the addSourceBuffer() original call or possibly a more recent changeType() call in the introspection result could assist interoperable debugging.

wolenetz:
This issue is part of a group of "introspection API" or similar API feature requests: #259, #209, #172, #40, #35.

PR #295, opened 2021-09-03
Use internal slots for various MediaSource and SourceBuffer object state

### LINKED ISSUE #286: Editorial: Modernize to use internal slots instead of "internal ... variable"
As @tidoust noted in a recent review comment (https://github.com/w3c/media-source/pull/282#discussion_r664101655), internal slots are how specs currently describe internal object variable state that is meaningful to the spec.

The existing MSE spec has a few of these which need updating (SourceBuffer append state is just one example).

wolenetz:
Note, track buffers and their internal state remain as "variables", not internal slots, in the spec for now. Later refactoring might make them internal slots too, though at lower priority.

This partially fixes #286

***
<a href="https://pr-preview.s3.amazonaws.com/wolenetz/media-source/pull/295.html" title="Last updated on Sep 3, 2021, 12:09 AM UTC (71c80b2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/media-source/295/6efee8e...wolenetz:71c80b2.html" title="Last updated on Sep 3, 2021, 12:09 AM UTC (71c80b2)">Diff</a>

### DIFF
--- a/media-source-respec.html
+++ b/media-source-respec.html
@@ -1263,7 +1263,7 @@ interface SourceBuffer : EventTarget {
 
           <ol class="method-algorithm">
             <li>Run the [=prepare append=] algorithm.</li>
-            <li>Add |data:BufferSource| to the end of the [=input buffer=].</li>
+            <li>Add |data:BufferSource| to the end of the {{SourceBuffer/[[input buffer]]}}.</li>
             <li>Set the {{SourceBuffer/updating}} attribute to true.</li>
             <li>[=Queue a task=] to [=fire an event=] named {{updatestart}} at this <a>SourceBuffer</a> object.</li>
             <li>Asynchronously run the [=buffer append=] algorithm.</li>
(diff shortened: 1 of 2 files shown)


wolenetz:
@tidoust / @mwatson2 - Am I correct that the current external reference by the BSF Registry of `generate timestamps flag` via the helpers in media-source.js would be more stable if instead that registry used xref to reference the media-source spec, and {{SourceBuffer/[[generate timestamps flag]]}} respec syntax to reference the definition of that internal slot? For now at least, I *think* I've updated media-source.js in this PR to use the updated fragment for this former variable that is now an internal slot.

Also, please review soon if you could, as rebasing this could get ugly since it touches quite a number of lines. Thanks!

tidoust (approved):
This all looks good to me! The need to write terms surrounded by `[[ ]]` and `{{ }}` in the source spec is not necessarily fantastic from an authoring perspective. Hopefully, the ability to better track and validate references to these terms makes it acceptable.

> @tidoust / @mwatson2 - Am I correct that the current external reference by the BSF Registry of generate timestamps flag via the helpers in media-source.js would be more stable if instead that registry used xref to reference the media-source spec, and {{SourceBuffer/[[generate timestamps flag]]}} respec syntax to reference the definition of that internal slot?

Yes and you were right to flag the definition of the internal slot with a `data-export` attribute so that it can be referenced by the BSF registry.

wolenetz:
@mwatson2 - please take a look, or is @tidoust's prior approval sufficient? Thanks!

mwatson2 (approved):
Looks good! The only thing I noticed - which exists in the pre-existing text and so is not a comment on this PR - is that some of these variables have explicitly stated types (boolean / TimeRanges) and some do not.

wolenetz:
> Looks good! The only thing I noticed - which exists in the pre-existing text and so is not a comment on this PR - is that some of these variables have explicitly stated types (boolean / TimeRanges) and some do not.

Thanks for the review. I too noticed some types missing. Some are simple, others like the "track buffer" concept have less strongly typed information, as it is left more to implementations. Possibly this is an area of improvement in the spec though, especially if things like buffered media introspection API is landed (where the notion of an app-visible "track buffer" object might make sense -- I've added a note around this to #259 ).

PR #300, opened 2021-09-27
Fix broken refs to HTML spec, adjust settings for FPWD

tidoust:
The spec referenced a number of terms in the HTML spec using links to the superseded HTML 5.1 spec. That spec actually redirects to the WHATWG HTML LS spec, which is good. Problem is a good chunk of anchors have been modified in the HTML spec and are no longer valid.

This update fixes all broken fragments using a combination of:
- "xref" terms, automatically handled by ReSpec, when possible (i.e. when the HTML spec exports them).
- fragment updates in media-source.js when the spec needs to reference non-exported terms. Longer-term, we'll need to work with HTML editors to make sure that the spec properly exports the terms that we need to reference.

These updates are notably needed to pass pubrules for publication as First Public Working Draft (I included them in the version I sent for publication to /TR).

One note on xref terms: the HTML spec no longer has the concept of "initial playback position". The concept seems to have been merged with "current playback position".

This update also includes a couple of other fixes to please pubrules:
- Required editor ID was missing for Mark.
- Link to Can I Use was semi-broken (they still managed to redirect the request but a more direct https://caniuse.com/mediasource can be used)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/media-source/pull/300.html" title="Last updated on Sep 29, 2021, 7:04 AM UTC (2182bb3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/media-source/300/09f30ae...2182bb3.html" title="Last updated on Sep 29, 2021, 7:04 AM UTC (2182bb3)">Diff</a>

### DIFF
--- a/media-source-respec.html
+++ b/media-source-respec.html
@@ -76,8 +76,6 @@
       // name of the WG
       group: "media",
 
-      implementationReportURI: "https://tidoust.github.io/media-source-testcoverage/",
-
       scheme: "https",
 
       github: "w3c/media-source",
--- a/media-source-respec.html
+++ b/media-source-respec.html
@@ -221,7 +221,7 @@
           <dd><p>The parent media source of a <a>SourceBuffer</a> object is the <a>MediaSource</a> object that created it.</p></dd>
 
           <dt><dfn id="presentation-start-time">Presentation Start Time</dfn></dt>
-          <dd><p>The presentation start time is the earliest time point in the presentation and specifies the <a def-id="videoref" name="current-playback-position">playback position</a> and <a def-id="videoref" name="earliest-possible-position">earliest possible position</a>. All presentations created using this specification have a presentation start time of 0.</p>
+          <dd><p>The presentation start time is the earliest time point in the presentation and specifies the initial <a def-id="videoref" name="current-playback-position">playback position</a> and <a def-id="videoref" name="earliest-possible-position">earliest possible position</a>. All presentations created using this specification have a presentation start time of 0.</p>
 
           <p class="note">For the purposes of determining if {{HTMLMediaElement}}.{{HTMLMediaElement/buffered}} contains a {{TimeRanges}} that includes the current playback position, implementations MAY choose to allow a current playback position at or after [=presentation start time=] and before the first {{TimeRanges}} to play the first {{TimeRanges}} if that {{TimeRanges}} starts within a reasonably short time, like 1 second, after [=presentation start time=]. This allowance accommodates the reality that muxed streams commonly do not begin all tracks precisely at [=presentation start time=]. Implementations MUST report the actual buffered range, regardless of this allowance.</p>
           </dd>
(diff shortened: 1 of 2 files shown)


tidoust:
I added a couple of updates to use a leveled shortname because [it turns out that it is required](https://lists.w3.org/Archives/Public/public-media-wg/2021Sep/0011.html) and to drop the link to the now outdated implementation report.

media-source-respec.html
```diff
@@ -224,7 +221,7 @@ <h3>Definitions</h3>
           <dd><p>The parent media source of a <a>SourceBuffer</a> object is the <a>MediaSource</a> object that created it.</p></dd>
 
           <dt><dfn id="presentation-start-time">Presentation Start Time</dfn></dt>
-          <dd><p>The presentation start time is the earliest time point in the presentation and specifies the <a def-id="videoref" name="initial-playback-position">initial playback position</a> and <a def-id="videoref" name="earliest-possible-position">earliest possible position</a>. All presentations created using this specification have a presentation start time of 0.</p>
+          <dd><p>The presentation start time is the earliest time point in the presentation and specifies the <a def-id="videoref" name="current-playback-position">playback position</a> and <a def-id="videoref" name="earliest-possible-position">earliest possible position</a>. All presentations created using this specification have a presentation start time of 0.</p>
```
wolenetz, line 224: current-playback-position changes during a playback. MSE prescribes both the earliest-possible-position and the initial-playback-position to both be 0. The linked HTMLMediaElement spec still references "initial playback position" in steps 10 and 11 of the "Once enough of the media data has been fetched to determine the duration of the media resource, its dimensions, and other metadata" in its media data processing steps list.

current-playback-position would be incorrect here. If we must update due a broken link for initial-playback-position, we could file an issue to get that concept exported from HTMLMediaElement for our use and meanwhile explicitly describe the linkage here as being the "initial playback position" for a MediaSource media resource as needed by the "media data processing steps" of HTMLMediaElement.

wolenetz (changes requested):
With the exception of the need to still refer to initial-playback-position somehow, the rest of this looks good to me.

media-source-respec.html
tidoust: Oops, I forgot to add "initial" back in the prose I changed but today I don't have anything better to offer than putting that adjective outside of the reference and referencing "current playback position", as in:
```html
the initial <a def-id="videoref" name="current-playback-position">playback position</a>
```

The steps 10 and 11 that you mention are the only places in HTML that mention "initial playback position". That concept as a whole no longer exists in HTML (it used to). That actually needs fixing in HTML, because "initial playback position" is wrapped in a `<var>` but that variable is not defined anywhere.

In other words, it's not only a matter of exporting a definition, it's a matter of re-introducing one that got dropped, or considering that "initial" is just common sense to mean the initial value of the "current playback position". I propose the latter for now until that gets improved in HTML.

tidoust: Filed an issue on the HTML spec at https://github.com/whatwg/html/issues/7129

wolenetz (approved):
Thank you for filing the HTML issue for initial-playback-position. This now LGTM.

PR #302, opened 2021-10-01
MSE-for-WebCodecs draft feature specification

wolenetz:
***
<a href="https://pr-preview.s3.amazonaws.com/wolenetz/media-source/pull/302.html" title="Last updated on Oct 1, 2021, 11:12 AM UTC (57e512d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/media-source/302/5135541...wolenetz:57e512d.html" title="Last updated on Oct 1, 2021, 11:12 AM UTC (57e512d)">Diff</a>

### DIFF
--- a/media-source-respec.html
+++ b/media-source-respec.html
@@ -118,7 +118,7 @@
 
     <link rel="stylesheet" href="mse.css">
   </head>
-  <body data-cite="html dom url">
+  <body data-cite="html dom url WEBCODECS">
     <section id="abstract">
       This specification extends {{HTMLMediaElement}} [[HTML]] to allow
       JavaScript to generate media streams for playback.
@@ -344,8 +344,9 @@ interface MediaSource : EventTarget {
 };
 
 dictionary SourceBufferConfig {
-  // Precisely one of these must be populated to signal via addSourceBuffer or
-  // changeType the intent to buffer the corresponding WebCodecs media.
+  // Precisely one of these WebCodecs config objects must be populated to
+  // signal via addSourceBuffer or changeType the intent to buffer the
+  // corresponding WebCodecs media.
   AudioDecoderConfig audioConfig;  // For appending EncodedAudioChunks
   VideoDecoderConfig videoConfig;  // For appending EncodedVideoChunks
 };
@@ -356,7 +357,6 @@ typedef (DOMString or SourceBufferConfig) TypeOrConfig;
 </pre>
       <section>
         <h3>Attributes</h3>
-        <p>REMOVE THIS PLACEHOLDER NOTE: TODO describe SourceBufferConfig and the updated addSourceBuffer and the linkage to [[WEBCODECS]]</p>
         <dl class="attributes" data-dfn-for="MediaSource"><dt><dfn><code>sourceBuffers</code></dfn> of type <span class="idlAttrType"><a>SourceBufferList</a></span>, readonly       </dt><dd>
           Contains the list of <a>SourceBuffer</a> objects associated with this <a>MediaSource</a>. When {{MediaSource/readyState}} equals {{ReadyState/""closed""}} this list will be empty. Once {{MediaSource/readyState}} transitions to {{ReadyState/""open""}} SourceBuffer objects can be added to this list by using {{MediaSource/addSourceBuffer()}}.
         </dd><dt><dfn><code>activeSourceBuffers</code></dfn> of type <span class="idlAttrType"><a>SourceBufferList</a></span>, readonly       </dt><dd>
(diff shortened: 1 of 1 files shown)


wolenetz:
@mwatson2, please review this full feature. There are probably some edge cases in pre-existing text that I missed that might still need some updating. Hopefully, the sequence of changes I put together in this PR enable easier piecemeal review.

@tidoust, please review for respec and standards-track aspects.

Thank you both!

media-source-respec.html
```diff
@@ -1128,17 +1222,28 @@ <h2><dfn>SourceBuffer</dfn> Object</h2>
...
-            of the timestamps in the media segment.
-          </p>
+};</pre><table class="simple" data-dfn-for="AppendMode"><tbody><tr><th colspan="2">Enumeration description</th></tr>
+      <tr><td><dfn><code id="idl-def-AppendMode.segments">segments</code></dfn></td><td>
+        <p>The timestamps in the media segment or {{EncodedChunks}} determine where the [=coded frames=] are placed in
+        the presentation. Media segments can be appended in any order. Discontinuous {{EncodedChunks}} should not be
+        appended in the same {{SourceBuffer/appendEncodedChunks()}} call: the application should instead wait for the
+        {{SourceBuffer}} to no longer be {{SourceBuffer/updating}}, and then call {{SourceBuffer/abort()}} or
+        {{SourceBuffer/changeType()}} before appending {{EncodedChunks}} that are discontinuous with those in the
+        previous {{SourceBuffer/appendEncodedChunks()}} call.</p>
```
mwatson2, line 1232: I'm curious why EncodedChunks cannot be appended in any order like media segments can ?

Does the above mean that to append, say, overlapping media (replacing some existing frames) we need to call `changeType` even if the type has not changed ?

media-source-respec.html
```diff
@@ -1155,13 +1260,30 @@ <h2><dfn>SourceBuffer</dfn> Object</h2>
...
+};
+
+// This typedef simplifies the syntax for describing either a single audio or
+// video chunk, or a sequence of chunks. While this typedef does not provide the
+// ability to enforce that all members of a sequence of chunks are the same kind
+// (audio versus video), the behavior of the appendEncodedChunks method enforces
+// that constraint: the implementation MUST reject a sequence containing both
+// audio and video chunks.
+typedef (sequence< (EncodedAudioChunk or EncodedVideoChunk) >
+            or EncodedAudioChunk or EncodedVideoChunk) EncodedChunks;
```
mwatson2, line 1281: Can't we write `sequence<EncodedAudioChunk> or sequence<EncodedVideoChunk> or EncodedAudioChunk or EncodedVideoChunk` ?

mwatson2 (approved):

media-source-respec.html
wolenetz: I agree this is not ergonomic/compatible with previous behavior, and should be improved.

If I recall from when I wrote this draft and the prototype in Chromium, sequences of out-of-order presentation time versus the decode sequence must still be supported. The issue, I think, arises from the lack of explicit "decode timestamp" in EncodedChunks, and the prototype/draft API instead assumes the order of chunk appends describes their decode sequence. There is no ability of the coded frame processing algorithm to detect discontinuities in EncodedChunk decode time, unlike bytestream-formatted media. Without some ability to determine actual discontinuity implicitly, explicit discontinuity signaling is needed (abort() or changeType()). Even so, `changeType` isn't required if type isn't changing, but there is indeed a discontinuity. Instead, abort() signals the discontinuity explicitly without changing type.

This could be improved/simplified for "segments" mode: keyframes themselves could be used as the implicit signal of potential discontinuity, especially since there is allowance for only 1 audio config or only 1 video config in a `SourceBuffer` configured to allow EncodedChunk appends (and therefore there is no need to have such per-keyframe discontinuity also impacting the processing of coded frames for any other track buffers in that `SourceBuffer`.)

I might be missing some other detail in my recollection of this old PR and prototype though. For example, "sequence" mode appends that are not discontinuous at keyframes, but might appear to be if they are from SAP-Type-2 encode sequences, could cause interop issues if keyframes are used as implicit potential discontinuities. Conversely, without implicit discontinuity signaling, lack of app explicitly signaling could impair reaching the intent of "sequence" mode buffering.

One solution might be to use keyframes as implicit potential discontinuity signaling, but disallow use of appending chunks in "sequence" mode.

media-source-respec.html
wolenetz: I believe I tried that originally in the prototype, but the IDL bindings generator failed. I fuzzily recall maybe it was due to an identifier length limit on the generated type name. Worth reconsidering.

wolenetz: I don't like leaving this with just a fuzzy recollection, so I checked the [Chromium prototype's commit description](https://source.chromium.org/chromium/chromium/src/+/6507c9cc4ae2d08d090d466da71741d8677380cf) for the change I landed that added this IDL:

>   Caveat: "sequence<A> or sequence<V>" cannot be disambiguated by the
>   bindings, so sequence<A or V> is used in this change. Regardless, the
>   eventual implementation would need to validate that all in the
>   sequence are either A or all are V (along with the usual validation
>   that appended chunks or frames also appear to use the most recent
>   SourceBufferConfig).

Likewise, my fuzzy recollection of identifier length was fuzzy. In that same change I *did* run into identifier length limits, but solved that by adding the long identifier into an exception list for that check in the generator:

>   Caveat: The bindings generator requires help when generated union type
>   identifiers are too long for some platforms. This change adds a
>   seventh case to the existing hard-coded lists of names that need
>   shortening with the generator.

chrisn:
There's a new [mse-for-webcodecs-draft branch](https://github.com/w3c/media-source/tree/mse-for-webcodecs-draft) in this repo that rebases the changes in this PR on top of the latest draft.

PR #303, opened 2021-10-04
Automate publication to /TR and handle static files

tidoust:
This update completes the job that validates and publishes the spec to handle static files as well (images, companion documents, MSE v1 errata) and to now publish Working Draft updates automatically to /TR.

The `ECHIDNA_TOKEN` secret was added to the repo settings.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,18 +1,48 @@
 name: CI
 on:
+  # Worflow runs on pull requests where it makes sure that the spec can be
+  # generated, that markup and IDL are valid, as well as on pushes to the
+  # default branch where it also deploys the generated spec to the gh-pages
+  # branch and publishes the result to /TR.
+  # The "workflow_dispatch" hook allows repo admins to trigger the workflow
+  # manually if needed.
   pull_request: {}
   push:
     branches: [main]
+  workflow_dispatch:
 
 jobs:
   main:
     name: Build, Validate and Deploy
     runs-on: ubuntu-20.04
     steps:
-      - uses: actions/checkout@v2
-      - uses: w3c/spec-prod@v2
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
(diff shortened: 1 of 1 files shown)


wolenetz (approved):
These in general look good to me, though I'm far less experienced with the CI configuration side of this. Thank you for doing this!

wolenetz:
@tidoust can this be merged?

wolenetz:
@tidoust Looking at previous, similar https://github.com/w3c/picture-in-picture/pull/200, especially the further pieces added to that to fix that spec's automatic publication, does this current change also need:
* Some modification to add a "Previous Version" link?
* Some modification to set the spec level? (IIUC, level might need to be "2", but I could be wrong of course).

wolenetz:
Hmm. I'm also uncertain how to unmark approval for this until I'm clear that my questions in https://github.com/w3c/media-source/pull/303#issuecomment-1042467586 are answered. For now, I've just self-requested a review, but really I'm pending @tidoust responding.

tidoust:
> Looking at previous, similar https://github.com/w3c/picture-in-picture/pull/200

The Picture-in-Picture spec uses Bikeshed. MSE uses ReSpec. ReSpec should not need a previous link and the shortname in the ReSpec config is already set to `media-source-2`. We didn't reflect the notion of level in the title in the First Public Working Draft, so no need to go beyond that (if that's something that feels needed, ReSpec's [`level`](https://respec.org/docs/#level) config parameter could be used but that is orthogonal to automation).

I'm afraid the only real way to test the publication workflow is to put it in production, so I'll take the liberty to go ahead and merge this PR. I'll fix parameters if publication fails.

tidoust:
There remained a parameter name hiccup in the job. Now fixed. I confirm that automatic publication is now up and running:
https://www.w3.org/TR/2022/WD-media-source-2-20220217/

wolenetz:
Ack. Thank you!

Issue #307, opened 2022-04-26
Editorial accessibility concerns and suggestions


matatk:
I'm co-chair of the Accessible Platform Architectures (APA) WG. We are currently reviewing your specification for any accessibility concerns. However this is a personal comment concentrating on editorial accessibility issues and suggestions for the document. As these are editorial in nature, I am filing them all under one issue; hope that's OK.

We can provide further info and suggestions if that would help. Also, we recognise that you may be aware of some of these already and, for example, be in the process of updating image descriptions.

Our technical review will follow. Here are the editorial accessibility concerns, with the sections they appear in (may not be exhaustive)...

* §1 - The new page describing the flowchart diagram (which 404s from the [current version of the spec](https://www.w3.org/TR/media-source-2/), but is present in the [Editor's Draft](https://w3c.github.io/media-source/)) is really good, though there are several small mismatches between the description and the diagram in its current version.

* §2, 4 - The enumeration description tables have two visually apparent columns, but only a single header that spans both columns. Consider either providing two separate column headers, or coding this as a definition list.

* §2.2, 4.2, 5.2 - Parameter definition tables
  - Please make the parameter names row headers, to make navigating the table easier for people who can't see the screen. Row headers will ensure that the parameter name gets announced whenever the person moves into a cell on a new row.
  - Consider centering the check and cross symbols, so it's easier to see which column they're in.
  - Consider putting horizontal lines between rows, to make it easier to follow the structure of the tables.

* §3.5.12 - The image representing the algorithm lacks a description. It's acknowledged that the image is itself a graphical representation of the algorithm described above it, but looks like it's a particularly concrete example, and it has specific features that should be described to help people interpret it. Another way to phrase this would be that the image isn't what I had in mind when reading the algorithm (though that may be due to lack of experience). For example, there are vertical lines and gradients at specific times in the image. It seems like a separate description may be warranted.

* §4, 6.1, 6.2, 7, 8, 9 - The "agenda" tag has very low contrast (it's black text on a dark red background). It doesn't seem that this will be in the final spec, but either way if it's possible to address it, that would be helpful.

Please let us know if you have any concerns and/or would like help addressing any of the above.

tidoust:
Thanks for the comments @matatk!

I prepared a couple of pull requests accordingly.

[...]

PR #310, opened 2022-05-05
Improve pipeline model text description clarity

wolenetz:
Fixes a typo (one of the flows was described incorrectly as going to a
Video Decoder, not an Audio Decoder), and fixes a couple omissions of
mentions of track buffer process indications between SourceBuffer 2
and Video Decoder, and between SourceBuffer 3 and Audio Decoder.

(I found these issues while investigating #307 and reviewing #308.)

### DIFF
--- a/pipeline_model_description.html
+++ b/pipeline_model_description.html
@@ -31,7 +31,6 @@ HTMLMediaElement.<p>
 
 
 
-
 <h3 id='sourcebuffer1'>Flowing down from SourceBuffer 1</h3>
 
 
@@ -46,11 +45,10 @@ labeled Video Tag Display Region.</p>
 then follows to an indicator of an open switch, which then continues to an
 indicator (green circle with an X) labeled Audio Device.</p>
 
-<p>The third Track Buffer also flows to a green box labeled Video Decoder,
+<p>The third Track Buffer also flows to a green box labeled Audio Decoder,
 which also flows to a switch, this time however indicated closed, which then
 also continues to the same indicator (green circle with an X) labeled Audio
 Device as the second Track Buffer.</p>
-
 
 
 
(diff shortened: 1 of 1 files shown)


wolenetz:
@mwatson2 - please review.

@tidoust - with #308 now merged to main, is there anything further needed to ensure both WD and ED have the updated text description once this PR is approved, squashed and merged to main?

tidoust:
> @tidoust - with https://github.com/w3c/media-source/pull/308 now merged to main, is there anything further needed to ensure both WD and ED have the updated text description once this PR is approved, squashed and merged to main?

Well, there seems to be a bug somewhere that crashes the update of the `gh-pages` branch :( I don't yet know what triggers it, perhaps an incompatibility between the part of the job that handles the static content and the part of the job that handles the spec itself. I will look into it. No need to block any of the pending pull requests on this, I'll re-generate the `gh-pages` branch once I have found and fixed the bug!

wolenetz:
>  I will look into it. No need to block any of the pending pull requests on this, I'll re-generate the gh-pages branch once I have found and fixed the bug!

Thank you. I'll await @mwatson2 review of this and eventually `gh-pages` will be right again :)

wolenetz:
@mwatson2 - please review this when you can. To clear the editorial PR backlog, I'm going to squash and merge this today though since it is indeed editorial.

pipeline_model_description.html
```diff
@@ -46,29 +45,35 @@ <h3 id='sourcebuffer1'>Flowing down from SourceBuffer 1</h3>
...
 
 <p>Flowing down from SourceBuffer 3 is a rose triangle with an arrow flowing to
-a green box labeled Audio Decoder, which then flows through a closed switch,
-which then also continues to the same indicator (green circle with an X)
-labeled Audio Device previously mentioned in SourceBuffer 1</p>
+a process indication labeled Track Buffer.<p>
+
+<p>This Track Buffer then flows to a green box labeled Audio Decoder, which
+then flows through a closed switch, which then also continues to the same
+indicator (green circle with an X) labeled Audio Device previously mentioned in
```
chrisn, line 75: I think we could rephrase this to describe the purpose of "indicator (green circle with an X)". In a flow chart (I just looked this up...) the symbol is a "summing junction" which suggests that any enabled audio tracks are summed, i.e., mixed to form the output audio. This corresponds with definition of `AudioTrack`. 

Unless you prefer to do it, I could raise a PR with suggested text along those lines.

wolenetz: Good catch. PR sounds good to me if you have time. I've filed an issue to ensure this gets done one way or the other :)
https://github.com/w3c/media-source/issues/314

PR #313, opened 2022-05-10
[CI job] Swap steps to fix publication to gh-pages

tidoust:
Publication to `gh-pages` currently fails because the w3c/spec-prod action cannot find the `.git/config` file when it looks for it. The action runs smoothly on a number of other repositories with very similar settings, and manages to publish the spec to /TR without any problem. One explanation could be that the step that copies static file somehow modifies the local state (and drops the `.git` folder).

This update swaps the steps to run the spec generation job first and deployment of static files last.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,19 +19,6 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v2
 
-      # If event is a push to main branch (and not a pull request),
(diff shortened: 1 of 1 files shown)


tidoust:
OK, I don't really know what triggers the problem but swapping the steps seems to work. Time permitting, I'll look into using another action to handle static files but propose to stick to that for now...

wolenetz:
Thanks for doing this!

PR #327, opened 2023-10-23
Editorial: remove redundant tables

marcoscaceres:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/media-source/pull/327.html" title="Last updated on Dec 21, 2023, 12:38 AM UTC (9f61f9a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/media-source/327/9f216ef...9f61f9a.html" title="Last updated on Dec 21, 2023, 12:38 AM UTC (9f61f9a)">Diff</a>

### DIFF
--- a/media-source-respec.html
+++ b/media-source-respec.html
@@ -571,6 +571,7 @@ interface MediaSource : EventTarget {
               {{removesourcebuffer}} at the <a>SourceBufferList</a> returned by {{MediaSource/sourceBuffers}}.</li>
             <li>Destroy all resources for |sourceBuffer|.</li>
           </ol>
+        </section>
 
         <h3><dfn>endOfStream()</dfn> method</h3>
           <p>Signals the end of the stream.</p>
--- a/media-source-respec.html
+++ b/media-source-respec.html
@@ -452,7 +452,6 @@ interface MediaSource : EventTarget {
             <li>Add the new object to {{MediaSource/sourceBuffers}} and [=queue a task=] to [=fire an event=] named {{addsourcebuffer}} at {{MediaSource/sourceBuffers}}.</li>
             <li>Return the new object.</li>
           </ol>
-        <table class="parameters"><tbody><tr><th>Parameter</th><th>Type</th><th>Nullable</th><th>Optional</th><th>Description</th></tr><tr><th class="prmName">|type|</th><td class="prmType">{{DOMString}}</td><td class="prmNullFalse"><span role="img" aria-label="False">✘</span></td><td class="prmOptFalse"><span role="img" aria-label="False">✘</span></td><td class="prmDesc"></td></tr></tbody></table><div><em>Return type: </em><a>SourceBuffer</a></div></dd>
 
           <dt><dfn><code>removeSourceBuffer</code></dfn></dt><dd>
           <p>Removes a {{SourceBuffer}} from {{MediaSource/sourceBuffers}}.</p>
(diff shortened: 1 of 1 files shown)


media-source-respec.html
```diff
@@ -622,11 +619,11 @@ <h3>Attributes</h3>
...
-
           <dt><dfn><code>setLiveSeekableRange</code></dfn></dt><dd>
-
+          <p>
+            The |start| of the range, in seconds measured from [=presentation start time=]. While set, and if {{MediaSource/duration}} equals positive Infinity, {{HTMLMediaElement}}.{{HTMLMediaElement/seekable}} will return a non-empty TimeRanges object with a lowest range start timestamp no greater than |start|.
+            The |end| of range, in seconds measured from [=presentation start time=]. While set, and if {{MediaSource/duration}} equals positive Infinity, {{HTMLMediaElement}}.{{HTMLMediaElement/seekable}} will return a non-empty TimeRanges object with a highest range end timestamp no less than |end|.
+          </p>
           <p>Updates {{MediaSource/[[live seekable range]]}} that is used in
             <a href="#htmlmediaelement-extensions">HTMLMediaElement Extensions</a> to modify
             {{HTMLMediaElement}}.{{HTMLMediaElement/seekable}} behavior.</p>
```
wolenetz, line 629: 1. The now-removed redundant parameter tables were interestingly reordered in the rendered spec to occur between the summary of the method and the method-algorithm steps. With this method having non-trivial parameters that still need description, manual relocation needs to occur... (e.g. swap the order of these two paragraphs).
2. For readability, the non-trivial parameters might need a better introduction here.
3. The two "While set, and if ...." portions summarize the normative behavior defined in #htmlmediaelement-extensions section, so they are also redundant with that section. However, it might be good to retain some redundancy of that here. They shouldn't be normative here in the way they were, since they directly referenced the MediaSource's `duration`, which is not necessarily what the current version of the seekable extension does (it conditions reference on a "recent" duration and a "recent live seekable range" to handle MSE-in-Workers.

Also (not in scope of this PR): We might want to consider extracting the self-contained, but probably non-standard, text about conditionally cross-context-communicating worker's live seekable range updates and duration updates out of the seekable attribute's extension text. WDYT?

I haven't proofed this in respec, but maybe this would be a good fix for now:

```suggestion
          <p>Updates {{MediaSource/[[live seekable range]]}} that is used in
            <a href="#htmlmediaelement-extensions">HTMLMediaElement Extensions</a> to modify
            {{HTMLMediaElement}}.{{HTMLMediaElement/seekable}} behavior.</p>
          <p>
            The |start| of the range, in seconds measured from [=presentation start time=].
            The |end| of range, in seconds measured from [=presentation start time=].
          </p>
          <p class="note">
            <a href="#htmlmediaelement-extensions">HTMLMediaElement Extensions</a> uses
            {{MediaSource/[[live seekable range]]}}, if not empty and if
            {{MediaSource/duration}} is positive infinity, to return a {{TimeRanges}}
            object with a lowest range start timestamp no greater than |start| and a highest
            range end timestamp no less than |end|. If this is a worker {{MediaSource}},
            cross-context communication latency will delay the effect of any duration or
            live seekable range updates on
            {{HTMLMediaElement}}.{{HTMLMediaElement/seekable}} behavior.
          </p>
         
```

media-source-respec.html
```diff
@@ -1468,7 +1435,7 @@ <h2><dfn>SourceBuffer</dfn> Object</h2>
             <li>[=Queue a task=] to [=fire an event=] named {{updatestart}} at this <a>SourceBuffer</a> object.</li>
             <li>Asynchronously run the [=buffer append=] algorithm.</li>
           </ol>
-        <table class="parameters"><tbody><tr><th>Parameter</th><th>Type</th><th>Nullable</th><th>Optional</th><th>Description</th></tr><tr><th class="prmName">|data|</th><td class="prmType">{{BufferSource}}</td><td class="prmNullFalse"><span role="img" aria-label="False">✘</span></td><td class="prmOptFalse"><span role="img" aria-label="False">✘</span></td><td class="prmDesc"></td></tr></tbody></table><div><em>Return type: </em>{{undefined}}</div></dd><dt><dfn><code>abort</code></dfn></dt><dd>
+
```
wolenetz, line 1438: Too much was removed. We need SourceBuffer.abort() here still. This might fix the issue:
```suggestion

        <dt><dfn><code>abort</code></dfn></dt><dd>

```

wolenetz (changes requested):
Thank you for putting this together!

media-source-respec.html
```diff
@@ -1468,7 +1435,9 @@ <h2><dfn>SourceBuffer</dfn> Object</h2>
...
+        <dt><dfn><code>abort</code></dfn></dt><dd>
```
marcoscaceres, line 1439:
```suggestion
        <dt><code><dfn>abort</dfn></code></dt><dd>
```

media-source-respec.html
marcoscaceres: > Also (not in scope of this PR): We might want to consider extracting the self-contained, but probably non-standard, text about conditionally cross-context-communicating worker's live seekable range updates and duration updates out of the seekable attribute's extension text. WDYT?

Sounds good. Yes.

marcoscaceres:
I'm going to do ahead and merge this so I can clean up the rest of the markup.

PR #328, opened 2023-10-23
Editorial: switch MediaSource to using headings

marcoscaceres:
So things are easier to find in the ToC. You can see the actual API shape now.  

Screenshot 2023-10-23 at 14 52 30

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/media-source/pull/328.html" title="Last updated on Dec 18, 2023, 10:39 PM UTC (ee00d1c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/media-source/328/2cb6b9a...ee00d1c.html" title="Last updated on Dec 18, 2023, 10:39 PM UTC (ee00d1c)">Diff</a>

### DIFF
--- a/media-source-respec.html
+++ b/media-source-respec.html
@@ -275,9 +275,9 @@
       </section>
     </section>
 
-    <section id="mediasource">
-      <h2><dfn>MediaSource</dfn> Object</h2>
-      <p>The MediaSource object represents a source of media data for an HTMLMediaElement. It keeps track of the {{MediaSource/readyState}} for this source as well as a list of <a>SourceBuffer</a> objects that can be used to add media data to the presentation. MediaSource objects are created by the web application and then attached to an HTMLMediaElement. The application uses the <a>SourceBuffer</a> objects in {{MediaSource/sourceBuffers}} to add media data to this source. The HTMLMediaElement fetches this media data from the <a>MediaSource</a> object when it is needed during playback.</p>
+    <section id="mediasource" data-dfn-for="MediaSource">
+      <h2><dfn>MediaSource</dfn> interface</h2>
+      <p>The {{MediaSource}} interface represents a source of media data for an {{HTMLMediaElement}}. It keeps track of the {{MediaSource/readyState}} for this source as well as a list of <a>SourceBuffer</a> objects that can be used to add media data to the presentation. MediaSource objects are created by the web application and then attached to an HTMLMediaElement. The application uses the <a>SourceBuffer</a> objects in {{MediaSource/sourceBuffers}} to add media data to this source. The HTMLMediaElement fetches this media data from the <a>MediaSource</a> object when it is needed during playback.</p>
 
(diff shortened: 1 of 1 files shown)


media-source-respec.html
```diff
@@ -404,22 +406,18 @@ <h3>Attributes</h3>
               <p class="note">{{SourceBuffer/appendBuffer()}} and {{MediaSource/endOfStream()}} can update the duration under certain circumstances.</p>
             </li>
           </ol>
-        </dd><dt><dfn><code>onsourceopen</code></dfn> of type {{EventHandler}}</dt><dd>
```
marcoscaceres, line 407: These are self describing as they are event handlers

wolenetz: Respec seems to not like these since they're now rendered in red in the WebIDL, probably due to no reference to or from defining text.

media-source-respec.html
```diff
@@ -344,11 +344,8 @@ <h2><dfn>MediaSource</dfn> Object</h2>
     undefined             clearLiveSeekableRange ();
     static boolean        isTypeSupported (DOMString type);
 };</pre>
-      <section>
-        <h3>Attributes</h3>
-        <dl class="attributes" data-dfn-for="MediaSource">
 
-          <dt><dfn><code>handle</code></dfn> of type <span class="idlAttrType"><a>MediaSourceHandle</a></span>, [SameObject] readonly</dt><dd>
+          <h3><dfn><code>handle</code></dfn> attribute</h3>
```
wolenetz, line 348: The outline raises MediaSource's algorithms to the same level as other interfaces. This seems incorrect to me. If the idea was to more precisely enumerate all of the various attributes and methods in the TOC, then consulting how similar was done for Algorithms (within each interface) seems a better route. Something (unproofed) like:

```suggestion
     <section>
        <h3>Attributes</h3>
        <dl class="attributes" data-dfn-for="MediaSource">
          <h4><dfn><code>handle</code></dfn> attribute</h3>
        ....
```

wolenetz (changes requested):

media-source-respec.html
marcoscaceres: Definitely want to get rid of all the `<dl>`s... let me check what happened with the algorithm level.

media-source-respec.html
marcoscaceres: Having a look... it should handle these just find as "self defining"

marcoscaceres:
@wolenetz, can you take another look. It should be good now, I think, but please confirm ✅

chrisn:
Looks good, I think we should so the same for `SourceBuffer` and the other interfaces.

marcoscaceres:
> I think we should so the same for SourceBuffer and the other interfaces. 

Yes, absolutely. I'll do them as a followup to keep the patch manageable.

jyavenard (approved):

Issue #320, opened 2023-05-29
Proposal: ManagedMediaSource API


jyavenard:
### Definitions

A “managed” MediaSource is one where more control over the MediaSource and its associated objects has been given over to the User Agent.

### Introduction

The explicit goal of the Media Source Extensions specification is to transfer more control over the streaming of media data from the User Agent to the application running in the page. This transfer of control can and has added points of inefficiencies, where the page does not have the same level of capabilities, knowledge, or even goals as the User Agent.
Examples of these inefficiencies include the management of buffer levels, the timing and amount of network access, and media variant selection. These inefficiencies have largely been immaterial on relatively powerful devices like modern general purpose computers. However, on devices with narrower capabilities, it can be difficult to achieve the same quality of playback with the MediaSource API as is possible with native playback paths provided by the User Agent.

The goal of the ManagedMediaSource API is to transfer some control back to the User Agent from the application for the purpose of increasing playback efficiency and performance, while retaining the ability for pages to control streaming of media data.

### Goals

* Make it easier for media web site authors to support streaming media playback on constrained capability devices.
* Allow User Agents to react to changes in available memory and networking capabilities.
* Reduce the power impact of MediaSource APIs.
* Efficiently support high-speed / high-energy-use networks (e.g., 5G cellular). 

#### Low-memory availability

A user loads a MSE-based website on a device with a limited amount of physical memory, and no ability to swap. The user plays some content on that website, and pauses that content. The system subsequently faces memory pressure, and requires applications (including the User Agent) to purge unused memory. If not enough memory is made available, those applications (including the User Agent) may be killed by the system in order to free up enough memory to perform the operation triggering this memory pressure.

The User Agent runs a version of the “Coded Frame Eviction” algorithm, removing ranges of buffered data in order to free memory for use by the system. At the end of this algorithm, the User Agent fires a “bufferedchange” event at every SourceBuffer affected by this algorithm, allowing the web application to be notified that it may need to re-request purged media data from the server before beginning playback.

#### Memory availability notification

When a call to appendBuffer is rejected with a QuotaExceededError exception, it can  indicate the amount of excessive bytes or time that caused the error.

#### Network Streaming changes

Currently, a web application is allowed to append media data into a Source Buffer at any time, up until that Source Buffer’s “buffer full flag” is set, indicating no additional data is allowed to be appended. However, a constrained device may want to coalesce network use into a small window, and allow the network to query for battery and bandwidth reasons. 

[...]

Issue #325, opened 2023-10-17
Redundancies, duplications, and general spec health


marcoscaceres:
The spec needs a significant cleanup. I know the spec is really old, but it seems to be using specification patterns we (W3C and WHATWG) abandoned about a decade ago. 

The terminology in the spec is also all over the place. It confuses objects, interfaces, etc. 

Task queuing is are also not done correctly throughout. 

The markup in the spec is also a bit of nightmare, making it extremely difficult for folks to contribute. 

It would be good if, like any software project, we put in place coding standards and use tooling (e.g., HTML Tidy or Prettier) to keep the markup sane. This can be done automatically on merge, as is done elsewhere at the W3C.

marcoscaceres:
Additionally, the sectioning of this spec could be done better. The method/attributes should actually be headings, not using `<dl><dt><dd>` patterns. 

The silly summary tables that replicate what the IDL does should go away too. They add limited value and duplicate what is already clearly stated in the WebIDL (and become a maintenance burden because they need to be kept in sync with the IDL).

wolenetz:
Additionally, we should consider reworking the https://en.wikipedia.org/wiki/COMEFROM spaghetti of `recent *` text that was originally added to make the MSE-in-Workers spec change more self-contained.

Examples:

- `recent element error`
- `recent duration` and `recent live seekable range` (I noted these when reviewing related spec PR today: https://github.com/w3c/media-source/pull/327#discussion_r1385666251)
- `recent intersection ranges`
There could be others that I missed in my brief search now.

Thanks to @padenot for [reminding ](https://github.com/w3c/media-source/pull/329#discussion_r1371336987) me about the COMEFROM antipattern/joke :)

wolenetz:
MSE took from HTMLMediaElement long ago the methodology for this, and it certainly needs updating. HTMLMediaElement still has occurrences of the synchronous set before queue task to fire the event. For example:
![image](https://github.com/w3c/media-source/assets/11578819/f50b4866-dab5-4afc-a9e3-bde2df503f88)

wolenetz:
Many MSE event handler on.... attributes need more precise specification per https://github.com/w3c/media-source/pull/329#discussion_r1371332674 :

Event handlers are specified like this:

[...]

PR #329, opened 2023-10-23
Add Managed Media Source Extensions API

jyavenard:
As proposed in #320.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/media-source/pull/329.html" title="Last updated on Dec 5, 2023, 11:27 PM UTC (12fc650)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/media-source/329/6f69e59...12fc650.html" title="Last updated on Dec 5, 2023, 11:27 PM UTC (12fc650)">Diff</a>


media-source-respec.html
```diff
@@ -3148,7 +3518,66 @@ <h2>Examples</h2>
...
+        });
+
+        const sourceBuffer = source.addSourceBuffer(mediaType);
+
+        // Set up the event handlers
+        sourceBuffer.onbufferedchange = (e) => {
+          console.log("onbufferedchange event fired.");
+          console.log(`Added Ranges: [${timeRangesToString(e.addedRanges)}]`);
+          console.log(`Removed Ranges: [${timeRangesToString(e.removedRanges)}]`);
+        };
```
marcoscaceres, line 3553:
```suggestion
        sourceBuffer.onbufferedchange = (e) => {
          console.log("onbufferedchange event fired.");
          console.log(`Added Ranges: ${timeRangesToString(e.addedRanges)}`);
          console.log(`Removed Ranges: ${timeRangesToString(e.removedRanges)}`);
        };
```

media-source-respec.html
```diff
@@ -2706,6 +2714,367 @@ <h3>Event Summary</h3>
       </section>
     </section>
 
+    <section data-cite="mimesniff" data-dfn-for="ManagedMediaSource">
+      <h2><dfn>ManagedMediaSource</dfn> interface</h2>
+      <p>
+        A {{ManagedMediaSource}} is a {{MediaSource}} that actively manages its
+        memory content. Unlike a {{MediaSource}}, the [=user agent=] MAY evict it
+        from a {{ManagedSourceBuffer}} when under [=ManagedMediaSource/memory
```
dalecurtis, line 2722: I think the key part is that the UA may evict memory at any time. E.g., under memory pressure.

media-source-respec.html
```diff
@@ -2706,6 +2714,367 @@ <h3>Event Summary</h3>
...
+
+      <pre class="idl">
+      [Exposed=(Window,DedicatedWorker)]
+      interface ManagedMediaSource : MediaSource {
+        constructor();
+        readonly attribute boolean streaming;
+        attribute EventHandler onstartstreaming;
+        attribute EventHandler onendstreaming;
+
+        static boolean isTypeSupported(DOMString type);
```
dalecurtis, line 2734: Does this not carry over automatically?

media-source-respec.html
```diff
@@ -2706,6 +2714,367 @@ <h3>Event Summary</h3>
...
+        {{SourceBuffer}} Monitoring algorithm is scheduled to run.
+      </p>
+      <p>
+        Having <dfn>enough managed data to ensure uninterrupted playback</dfn>
+        is an implementation specific condition where the user agent determines that
+        it currently has enough data to play the presentation without stalling for a
+        meaningful period of time and that it can achieve a balance between optimal
+        memory and energy usage. This condition is constantly evaluated to determine
+        when to transition the media element into and out of the
+        {{HTMLMediaElement/HAVE_ENOUGH_DATA}} ready state. These transitions
```
dalecurtis, line 2846: Changing HAVE_ENOUGH_DATA semantics seems a bit risky, but likely fine for a new API like this -- though I suspect many clients will just blindly change MSE to MMSE. I was assuming all new state changes would come through the new events and readyState would be unchanged relative to normal MSE.

dalecurtis (commented):
Defer to editors for approval, but largely looks good to me.

media-source-respec.html
jyavenard: It would. However being a static it requires the interface type to be defined. 
on iPhone, the MediaSource interface itself isn’t defined and you can’t override a static method. It was also considered that the range of what could be supported in Managed MSE not always be identical to MSE. 

And as such, the interface method was duplicated.

media-source-respec.html
jyavenard: Are you suggesting a change here?

media-source-respec.html
jyavenard: This is mostly a spec verbiage. I found it easier to reuse the verbiage of the MSE.

readyState of the video element is indeed unchanged with MMS. But it does drive on when the streaming event will be fired

media-source-respec.html
```diff
@@ -2706,6 +2714,367 @@ <h3>Event Summary</h3>
       </section>
     </section>
 
+    <section data-cite="mimesniff" data-dfn-for="ManagedMediaSource">
+      <h2><dfn>ManagedMediaSource</dfn> interface</h2>
+      <p>
+        A {{ManagedMediaSource}} is a {{MediaSource}} that actively manages its
+        memory content. Unlike a {{MediaSource}}, the [=user agent=] MAY evict it
```
chrisn, line 2721: It's not quite clear what the "it" here is referring to, so I'd suggest a phasing such as:

> the user agent MAY evict data from a ManagedSourceBuffer when under memory pressure

media-source-respec.html
dalecurtis: Yes, I think the normative text should effectively say the User Agent may evict memory for any reason, while the non-normative section can give examples like memory pressure, etc.

media-source-respec.html
dalecurtis: Hmm, that seems like it just introduces more spec text than necessary. I'd probably drop this section unless you really need it then.

media-source-respec.html
dalecurtis: Probably this text should just refer to the streaming event then rather than readyState, but defer to spec editors.

media-source-respec.html
jyavenard: I've removed the method entirely

media-source-respec.html
jyavenard: I re-wrote it as:
```
        A {{ManagedMediaSource}} is a {{MediaSource}} that actively manages its
        memory content. Unlike a {{MediaSource}}, the [=user agent=] MAY evict evict content
        from its {{MediaSource/sourceBuffers}} (populated with {{ManagedSourceBuffer}}) for any reasons, including when under [=ManagedMediaSource/memory
        pressure=].

```

chrisn: One minor nit, "reasons" should be "reason", but otherwise this looks good to me

media-source-respec.html
```diff
@@ -2706,6 +2714,317 @@ <h3>Event Summary</h3>
...
+          {{SourceBuffer/buffered}} attribute before the changes occurred.
+        </li>
+        <li>
+          Let |new buffered ranges:normalized TimeRanges| equal the new
+          {{SourceBuffer/buffered}} {{TimeRanges}}.
+        </li>
+        <li>
+          Let |added:normalized TimeRanges| equal the |previous buffered
+          ranges:normalized TimeRanges| subtracted from |new buffered
+          ranges:normalized TimeRanges|.
```
padenot, line 2971: Do we have `TimeRange` arithmetic defined somewhere?

media-source-respec.html
```diff
@@ -2706,6 +2714,317 @@ <h3>Event Summary</h3>
...
+          ranges:normalized TimeRanges| subtracted from |previous buffered
+          ranges:normalized TimeRanges|.
+        </li>
+        <li>
+          Let |eventInitDict| be a new {{BufferedChangeEventInit}} dictionary
+          initialized with |added| as its {{BufferedChangeEventInit/addedRanges}} and
+          |removed| as its {{BufferedChangeEventInit/removedRanges}}
+        </li>
+        <li>
+          [=Queue a task=] to [=fire an event=] named {{bufferedchange}} at |buffer|
```
padenot, line 2984: Use the prose above to fire events, it's what we should do these days. This formulation, despite being used everything, is imprecise. See https://github.com/whatwg/html/issues/4980 for more background on this issue.

media-source-respec.html
```diff
@@ -2706,6 +2714,317 @@ <h3>Event Summary</h3>
...
+            </dd>
+          </dl>
+        </li>
+        <li>
+          Let |removal ranges:normalized TimeRanges| equal a list of presentation
+          time ranges that can be evicted from the presentation to ensure
+          uninterrupted playback from {{HTMLMediaElement.currentTime}} until such
+          presentation could be fetched again.
+          <p class="note">
+            Implementations can use different methods for selecting |removal ranges|
```
padenot, line 3014: "strategies" is a better fit for what we're attempting to convey here.

media-source-respec.html
```diff
@@ -2706,6 +2714,317 @@ <h3>Event Summary</h3>
...
+            so web applications shouldn't depend on a specific behavior. The web
```
padenot, line 3015: Use of RFC2119 in non-normative part of a specification (such as "Note:", etc.) is forbidden, so we need to rephrase this.

media-source-respec.html
```diff
@@ -3148,7 +3468,66 @@ <h2>Examples</h2>
...
+        // Check if the type of video format / codec is supported.
+        if (ManagedMediaSource?.isTypeSupported(mediaType)) {
+          return; // Not supported, do something else.
+        }
+
+        // Set up video and its managed source.
+        const video = document.createElement("video");
+        const source = new ManagedMediaSource();
+
+        video.disableRemotePlayback = true;
```
padenot, line 3487: Unrelated?

media-source-respec.html
```diff
@@ -412,28 +418,30 @@ <h3>Attributes</h3>
...
-              </dl>
             </li>
-            <li>Add the new object to {{MediaSource/sourceBuffers}} and [=queue a task=] to [=fire an event=] named {{addsourcebuffer}} at {{MediaSource/sourceBuffers}}.</li>
-            <li>Return the new object.</li>
+            <li>
+              If |buffer|'s {{SourceBuffer/[[generate timestamps flag]]}} is true, set |buffer|'s {{SourceBuffer/mode}} to {{AppendMode/"sequence"}}.
+              Otherwise, set |buffer|'s {{SourceBuffer/mode}} to {{AppendMode/"segments"}}.
+            </li>
+            <li>[=List/Append=] |buffer| to |this|'s {{MediaSource/sourceBuffers}}.</li>
+            <li>[=Queue a task=] to [=fire an event=] named {{addsourcebuffer}} at |this|'s {{MediaSource/sourceBuffers}}.</li>
```
padenot, line 442: See the comment about this type of prose below, let's make this correct while we're here.

media-source-respec.html
```diff
@@ -273,7 +273,13 @@ <h2><dfn>MediaSource</dfn> Object</h2>
         a media element=] algorithm. The extended [=resource fetch algorithm=] uses this internal slot to conditionally
         fail attachment of a {{MediaSource}} using a {{MediaSourceHandle}} set on a {{HTMLMediaElement}}'s
         {{HTMLMediaElement/srcObject}} attribute.</p>
-
+      <p>
+        Each {{MediaSource}} object has a
+        <dfn data-dfn-for="MediaSource">[[\is actively managed flag]]</dfn> internal
+        slot that stores a boolean indicating if the {{MediaSource}} object is
+        actively managing its content or not. It is initialized to false when the
```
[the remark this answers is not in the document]
padenot: Also, I can't find where it's set to true, but it's best to use the type anyway.

padenot (commented):
Lots of editorial comments, but the general concept looks good to me still.

media-source-respec.html
```diff
@@ -2706,6 +2714,317 @@ <h3>Event Summary</h3>
       </section>
     </section>
 
+    <section data-cite="mimesniff" data-dfn-for="ManagedMediaSource">
+      <h2><dfn>ManagedMediaSource</dfn> interface</h2>
+      <p>
+        A {{ManagedMediaSource}} is a {{MediaSource}} that actively manages its
+        memory content. Unlike a {{MediaSource}}, the [=user agent=] MAY evict evict content
```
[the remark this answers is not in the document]
jyavenard: Well, that MediaSource can almost be used inter-changeably with ManagedMediaSource is the intent. 
the memory pressure condition itself is optional. There may never be any if the UA doesn’t implement it

media-source-respec.html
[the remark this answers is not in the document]
jyavenard: I was aiming to not have MediaSource definition refers to the type ManagedMediaSource. Those cycled dependencies are a pain to deal with in code, I figure avoiding that in the spec itself would be preferred

media-source-respec.html
```diff
@@ -273,7 +273,13 @@ <h2><dfn>MediaSource</dfn> Object</h2>
         a media element=] algorithm. The extended [=resource fetch algorithm=] uses this internal slot to conditionally
         fail attachment of a {{MediaSource}} using a {{MediaSourceHandle}} set on a {{HTMLMediaElement}}'s
         {{HTMLMediaElement/srcObject}} attribute.</p>
-
+      <p>
+        Each {{MediaSource}} object has a
+        <dfn data-dfn-for="MediaSource">[[\is actively managed flag]]</dfn> internal
```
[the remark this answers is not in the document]
jyavenard: I used the naming style used elsewhere in those spec. 
i understand it’s not proper, but I think consistency within the spec is likely preferred

media-source-respec.html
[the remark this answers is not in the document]
jyavenard: > Also, I can't find where it's set to true, but it's best to use the type anyway.

yeah, it used to be set in the ManagedMediaSource definition, but that got accidentally removed in the later fix-up

media-source-respec.html
jyavenard: this is existing verbiage already used earlier. So if this is changed it would be better done in a separate PR.

media-source-respec.html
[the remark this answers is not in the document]
jyavenard: I've changed as hinted.

media-source-respec.html
```diff
@@ -2706,6 +2714,317 @@ <h3>Event Summary</h3>
...
+          <tr>
+            <td>
+              <dfn class="event">bufferedchange</dfn>
+            </td>
+            <td>{{Event}}</td>
+            <td>
+              The {{ManagedSourceBuffer}}'s buffered range changed following a
+              call to {{SourceBuffer/appendBuffer()}}, {{SourceBuffer/remove()}},
+              or as a consequence of the user agent running the
+              [=ManagedSourceBuffer/memory pressure=] algorithm.
```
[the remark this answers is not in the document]
jyavenard: That is the intent yes. The ManagedSourceBuffer can evict content at any time, for any reason suitable to the User Agent.

I'll amend the text accordingly.

media-source-respec.html
chrisn: So the advice to developers would be to use Media Capabilities? (which sounds like a good thing). It could be worth adding an informative note to say that.

jyavenard: No. You use MediaSource.isTypeSupported() from the base class.
There's no longer a distinction between what ManagedMediaSource can support and MediaSource

media-source-respec.html
```diff
@@ -2706,6 +2714,317 @@ <h3>Event Summary</h3>
...
+            </dd>
+          </dl>
+        </li>
+        <li>Run the {{MediaSource}} [=MediaSource/SourceBuffer Monitoring=] algorithm.</li>
+        <li>
+          <dl class="switch">
+            <dt>
+              If {{HTMLMediaElement}}'s {{HTMLMediaElement/buffered}} attribute
+              contains a {{TimeRanges}} that includes the current playback position
+              and not [=enough managed data to ensure uninterrupted playback=]:
```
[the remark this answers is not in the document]
jyavenard: The current text of this PR re-use the wording of the existing SourceBuffer monitoring algorithm ; I feel like we would lose the consistency between various parts of this spec as a consequence of making it proper just here.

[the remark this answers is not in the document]
jyavenard: ok, after online discussion. I will do so.

the entire gestion of events is completely broken in this spec (and to some extents in the HTML media part too).

Everywhere the attribute is modified first and then an event is queued to the element. Meaning that the attribute could have a different value at the time the event  is read in the event handler

media-source-respec.html
```diff
@@ -2706,6 +2714,317 @@ <h3>Event Summary</h3>
...
+            </td>
+          </tr>
+        </tbody>
+      </table>
+
+      <h3>Algorithms</h3>
+      <h4>Buffered Change</h4>
+      <p>
+        The following steps are run at the completion of all operations to the {{ManagedSourceBuffer}}
+        |buffer:ManagedSourceBuffer| that would cause a |buffer|'s {{SourceBuffer/buffered}} to change.
```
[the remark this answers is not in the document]
jyavenard: I'll do so in a follow-up

media-source-respec.html
```diff
@@ -2706,6 +2714,317 @@ <h3>Event Summary</h3>
...
+        That is once {{SourceBuffer/appendBuffer()}}, {{SourceBuffer/remove()}} or
+        [=ManagedSourceBuffer/memory pressure=] algorithm have completed.
+      </p>
+      <ol>
+        <li>
+          Let |previous buffered ranges:normalized TimeRanges| equal the
+          {{SourceBuffer/buffered}} attribute before the changes occurred.
```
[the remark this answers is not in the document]
jyavenard: I prefer to not touch anything in the MediaSource spec at this stage. I will come back to it

media-source-respec.html
jyavenard: To rework this so that it fits within the existing (broken) event handling requires a rather intensive rework of the spec. 

@marcoscaceres is going to revisit it all part of #325

media-source-respec.html
marcoscaceres: @padenot, the events queuing is broken throughout the entire spec... but let's def fix as a follow up.

media-source-respec.html
```diff
@@ -922,6 +913,7 @@ <h4><dfn>Detaching from a media element</dfn></h4>
               </dl></li>
             <li>Set {{MediaSource/[[port to main]]}} null.</li>
             <li>Set the {{MediaSource/readyState}} attribute to {{ReadyState/""closed""}}.</li>
+            <li>If [=this=] is a {{ManagedMediaSource}} set {{ManagedMediaSource/streaming}} attribute to `false`</li>
```
marcoscaceres, line 916:
```suggestion
            <li>If [=this=] is a {{ManagedMediaSource}} set {{ManagedMediaSource/streaming}} attribute to `false`.</li>
```

media-source-respec.html
```diff
@@ -2719,6 +2711,280 @@ <h3>Event Summary</h3>
...
+      </p>
+      <p>
+        Being <dfn>able to retrieve and buffer data in an efficient way</dfn>
+        is an implementation defined condition where the user agent determines that
+        it can fetch new data in an energy efficient manner while able to achieve the desired memory usage.
+      </p>
+      <ol class="algorithm">
+        <li>Run the {{MediaSource}} [=MediaSource/SourceBuffer Monitoring=] algorithm.</li>
+        <li>Let |can play uninterupted and efficiently| be a flag that is true if the {{HTMLMediaElement/buffered}} attribute contains a {{TimeRanges}} that includes the current playback position and [=enough managed data to ensure uninterrupted playback=]
+          and [=able to retrieve and buffer data in an efficient way=]
```
marcoscaceres, line 2800:
```suggestion
          and is [=able to retrieve and buffer data in an efficient way=]
```

media-source-respec.html
jyavenard: Just what Firefox’s TimeInterval arithmetic is doing :)

media-source-respec.html
```diff
@@ -2706,6 +2714,317 @@ <h3>Event Summary</h3>
...
+      <dl class="attributes" data-dfn-for="ManagedMediaSource">
+        <dt>
+          <dfn>streaming</dfn>
+        </dt>
+        <dd>
+          <p>On getting:</p>
+          <ol>
+            <li>
+              If [=this=]'s {{MediaSource/readyState}} attribute is
+              {{ReadyState/"closed"}} then return `false`.
```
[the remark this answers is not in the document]
jyavenard: Done

media-source-respec.html
```diff
@@ -2719,6 +2711,280 @@ <h3>Event Summary</h3>
...
+          </tr>
+        </thead>
+        <tbody>
+          <tr>
+            <td>
+              <dfn class="event">startstreaming</dfn>
+            </td>
+            <td>{{Event}}</td>
+            <td>
+              {{ManagedMediaSource/streaming}} attribute changed from `false` to `true`.
```
marcoscaceres, line 2763:
```suggestion
              A {{ManagedMediaSource}}'s {{ManagedMediaSource/streaming}} attribute changed from `false` to `true`.
```

media-source-respec.html
```diff
@@ -2719,6 +2711,280 @@ <h3>Event Summary</h3>
...
+            </td>
+          </tr>
+          <tr>
+            <td>
+              <dfn class="event">endstreaming</dfn>
+            </td>
+            <td>{{Event}}</td>
+            <td>
+              {{ManagedMediaSource/streaming}} attribute changed from `true` to `false`.
```
marcoscaceres, line 2772:
```suggestion
              A {{ManagedMediaSource}}'s {{ManagedMediaSource/streaming}} attribute changed from `true` to `false`.
```

media-source-respec.html
jyavenard: done, reworked it all.

media-source-respec.html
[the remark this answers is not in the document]
wolenetz: MSE took from HTMLMediaElement long ago the methodology for this, and it certainly needs updating. HTMLMediaElement still has occurrences of the synchronous set before queue task to fire the event. For example:
image

media-source-respec.html
```diff
@@ -2719,6 +2711,280 @@ <h3>Event Summary</h3>
...
+
+    <section data-dfn-for="BufferedChangeEvent">
+      <h2><dfn>BufferedChangeEvent</dfn> interface</h2>
+      <pre class="idl">
+        [Exposed=(Window,DedicatedWorker)]
+        interface BufferedChangeEvent : Event {
+          constructor(DOMString type, BufferedChangeEventInit eventInitDict);
+
+          [SameObject] readonly attribute TimeRanges addedRanges;
+          [SameObject] readonly attribute TimeRanges removedRanges;
```
wolenetz, line 2842: It's not obvious how an app might differentiate bufferedchangeevent's removedRanges from being caused by memory cleanup versus explicit remove that happened near in time to the memory cleanup (or other reasons like re-opening the ManagedMediaSource after it previously had been in "ended" readyState.) Is such differentiation desired by MMSE apps? If such differentiation were needed later, would this API shape be flexible to provide that information to the app?

media-source-respec.html
```diff
@@ -2719,6 +2711,280 @@ <h3>Event Summary</h3>
...
+        </dt>
+        <dd>
+          The time ranges added during the last run of the [=coded frame processing=] algorithm.
+        </dd>
+        <dt>
+          <dfn>removedRanges</dfn>
+        </dt>
+        <dd>
+          The time ranges removed during the last run of the [=coded frame removal=] or [=coded frame eviction=] algorithm or if the
+          user agent evicted content in response to a [=ManagedSourceBuffer/memory cleanup=].
```
wolenetz, line 2863: nit: "during the last run" is slightly confusing. [SameObject] implies "the last run" changes (which it should), but "during the last run" could be misinterpreted as whatever the most recent of these runs is whenever those attributes are queried.

media-source-respec.html
```diff
@@ -2719,6 +2711,280 @@ <h3>Event Summary</h3>
...
+          <tr>
+            <td>
+              <dfn class="event">bufferedchange</dfn>
+            </td>
+            <td>{{Event}}</td>
+            <td>
+              The {{ManagedSourceBuffer}}'s buffered range changed following a
+              call to {{SourceBuffer/appendBuffer()}}, {{SourceBuffer/remove()}},
+              or as a consequence of the user agent running the
+              [=ManagedSourceBuffer/memory cleanup=] algorithm.
```
wolenetz, line 2905: Here, in the BufferedChangeEvent's addedRanges and removedRanges, and in the "Buffered Change" algorithm, below, effective change of ManagedSourceBuffer's buffered attribute can come from other operations not mentioned.
Beyond appendBuffer, remove and memory cleanup, a ManagedSourceBuffer's buffered range additions and removals can be caused by open->ended MediaSource readyState transition (on endOfStream(<no error>)) and from ended->open transition, respectively: intersection ranges logic in the calculation of the SourceBuffer buffered attribute is conditioned on readyState: if ended, then "jagged ended" muxed tracks' highest end time is used for all those tracks.

media-source-respec.html
```diff
@@ -3161,7 +3428,65 @@ <h2>Examples</h2>
...
     </section>
+    <h3>Use of the Managed Media Source Extensions</h3>
+    <pre class="example html">
+      &lt;script&gt;
+      async function setUpVideoStream() {
+        // Specific video format and codec
+        const mediaType = 'video/mp4; codecs="mp4a.40.2,avc1.4d4015"';
+
+        // Check if the type of video format / codec is supported.
+        if (ManagedMediaSource?.isTypeSupported(mediaType)) {
```
wolenetz, line 3439: Unless I'm seriously mistaken, it looks like the result of this "if" statement is the opposite of what is wanted:
It currently is like, "if ManagedMediaSource is defined and has a method isTypeSupported that, when called with mediaType, returns true, then fail."
aside: nice optional chaining usage, though :)

wolenetz (changes requested):
About half nits. Lots of comments. Looks mostly good to me though!

media-source-respec.html
MattiasBuelens: > aside: nice optional chaining usage, though :)

Unfortunately, optional chaining doesn't work here... If a global variable named `ManagedMediaSource` doesn't exist, then this will still throw a `ReferenceError`:
image

wolenetz: @MattiasBuelens You're correct. Maybe instead:
if (!this.ManagedMediaSource??.isTypeSupported(mediaType)) { /* Either MMSE itself isn't supported or MMSE is supported but doesn't support mediaType */ }

media-source-respec.html
```diff
@@ -102,7 +102,7 @@
 
     <link rel="stylesheet" href="mse.css">
   </head>
-  <body data-cite="html dom url fileapi">
+  <body data-cite="html dom url fileapi infra">
     <section id="abstract">
       This specification extends {{HTMLMediaElement}} [[HTML]] to allow
       JavaScript to generate media streams for playback.
```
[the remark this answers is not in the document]
marcoscaceres: @wolenetz, can you clarify a bit more what you mean?

media-source-respec.html
```diff
@@ -425,28 +423,21 @@ <h3>Attributes</h3>
...
-              </dl>
             </li>
-            <li>Add the new object to {{MediaSource/sourceBuffers}} and [=queue a task=] to [=fire an event=] named {{addsourcebuffer}} at {{MediaSource/sourceBuffers}}.</li>
-            <li>Return the new object.</li>
+            <li>
+              If |buffer|'s {{SourceBuffer/[[generate timestamps flag]]}} is true, set |buffer|'s {{SourceBuffer/mode}} to {{AppendMode/"sequence"}}.
+              Otherwise, set |buffer|'s {{SourceBuffer/mode}} to {{AppendMode/"segments"}}.
+            </li>
+            <li>[=List/Append=] |buffer| to |this|'s {{MediaSource/sourceBuffers}}.</li>
+            <li>[=Queue a task=] to [=fire an event=] named {{addsourcebuffer}} at |this|'s {{MediaSource/sourceBuffers}}.</li>
```
marcoscaceres, line 438:
```suggestion
            <li>[=Queue a task=] to [=fire an event=] named {{addsourcebuffer}} at [=this=]'s {{MediaSource/sourceBuffers}}.</li>
```

media-source-respec.html
```diff
@@ -425,28 +423,21 @@ <h3>Attributes</h3>
               </p>
             </li>
             <li>If the {{MediaSource/readyState}} attribute is not in the {{ReadyState/""open""}} state then throw an {{InvalidStateError}} exception and abort these steps.</li>
-            <li>Create a new <a>SourceBuffer</a> object and associated resources.</li>
-            <li>Set the {{SourceBuffer/[[generate timestamps flag]]}} on the new object to the value in the "Generate
-              Timestamps Flag" column of the byte stream format registry [[MSE-REGISTRY]] entry that is associated with
+            <li>
+              Let |buffer| be a new instance of a {{ManagedSourceBuffer}} if |this| is a {{ManagedMediaSource}}, or a {{SourceBuffer}} otherwise, with their respective associated resources.            </li>
```
marcoscaceres, line 427:
```suggestion
              Let |buffer| be a new instance of a {{ManagedSourceBuffer}} if [=this=] is a {{ManagedMediaSource}}, or a {{SourceBuffer}} otherwise, with their respective associated resources. </li>
```

media-source-respec.html
```diff
@@ -425,28 +423,21 @@ <h3>Attributes</h3>
...
-                </dd>
-              </dl>
             </li>
-            <li>Add the new object to {{MediaSource/sourceBuffers}} and [=queue a task=] to [=fire an event=] named {{addsourcebuffer}} at {{MediaSource/sourceBuffers}}.</li>
-            <li>Return the new object.</li>
+            <li>
+              If |buffer|'s {{SourceBuffer/[[generate timestamps flag]]}} is true, set |buffer|'s {{SourceBuffer/mode}} to {{AppendMode/"sequence"}}.
+              Otherwise, set |buffer|'s {{SourceBuffer/mode}} to {{AppendMode/"segments"}}.
+            </li>
+            <li>[=List/Append=] |buffer| to |this|'s {{MediaSource/sourceBuffers}}.</li>
```
marcoscaceres, line 437:
```suggestion
            <li>[=List/Append=] |buffer| to [=this=]'s {{MediaSource/sourceBuffers}}.</li>
```

media-source-respec.html
```diff
@@ -2719,6 +2711,280 @@ <h3>Event Summary</h3>
...
+        <li>Run the {{MediaSource}} [=MediaSource/SourceBuffer Monitoring=] algorithm.</li>
+        <li>Let |can play uninterupted and efficiently| be a flag that is true if the {{HTMLMediaElement/buffered}} attribute contains a {{TimeRanges}} that includes the current playback position and [=enough managed data to ensure uninterrupted playback=]
+          and is [=able to retrieve and buffer data in an efficient way=]
+          <dl class="switch">
+            <dt>If |can play uninterupted and efficiently| is not equal to {{ManagedMediaSource/streaming}}, [=queue an element task=] on the [=media element=]
+              that runs the following steps:
+            </dt>
+            <dd>
+              <ol>
+                <li>Set |this| {{ManagedMediaSource/streaming}} attribute to |can play uninterupted and efficiently|.</li>
```
marcoscaceres, line 2807:
```suggestion
                <li>Set [=this=] {{ManagedMediaSource/streaming}} attribute to |can play uninterupted and efficiently|.</li>
```

media-source-respec.html
```diff
@@ -2719,6 +2711,280 @@ <h3>Event Summary</h3>
...
+        for any reason.
+      </p>
+
+      <pre class="idl">
+      [Exposed=(Window,DedicatedWorker)]
+      interface ManagedMediaSource : MediaSource {
+        constructor();
+        readonly attribute boolean streaming;
+        attribute EventHandler onstartstreaming;
+        attribute EventHandler onendstreaming;
```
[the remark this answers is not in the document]
marcoscaceres: I'll take a look. These are "self defining", as they are event handlers so don't need definitions.

media-source-respec.html
```diff
@@ -2719,6 +2711,280 @@ <h3>Event Summary</h3>
...
+          and is [=able to retrieve and buffer data in an efficient way=]
+          <dl class="switch">
+            <dt>If |can play uninterrupted and efficiently| is not equal to {{ManagedMediaSource/streaming}}, [=queue an element task=] on the [=media element=]
+              that runs the following steps:
+            </dt>
+            <dd>
+              <ol>
+                <li>Set [=this=] {{ManagedMediaSource/streaming}} attribute to |can play uninterrupted and efficiently|.</li>
+                <li>If |can play uninterrupted and efficiently| is false [=fire an event=] called {{startstreaming}} at the {{ManagedMediaSource}}.</li>
+                <li>Else [=fire an event=] called {{endstreaming}} at the {{ManagedMediaSource}}.</li>
```
marcoscaceres, line 2809:
```suggestion
                <li>Otherwise, [=fire an event=] called {{endstreaming}} at the {{ManagedMediaSource}}.</li>
```

media-source-respec.html
```diff
@@ -2719,6 +2711,280 @@ <h3>Event Summary</h3>
...
+        interface BufferedChangeEvent : Event {
+          constructor(DOMString type, BufferedChangeEventInit eventInitDict);
+
+          [SameObject] readonly attribute TimeRanges addedRanges;
+          [SameObject] readonly attribute TimeRanges removedRanges;
+        };
+
+        dictionary BufferedChangeEventInit : EventInit {
+          TimeRanges addedRanges;
+          TimeRanges removedRanges;
```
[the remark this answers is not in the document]
marcoscaceres: These should also be ok as self defining... will take a look.

media-source-respec.html
marcoscaceres, line 3439:
```suggestion
        if (window.ManagedMediaSource?.isTypeSupported(mediaType)) {
```

marcoscaceres: I went with `window`

media-source-respec.html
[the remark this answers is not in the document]
wolenetz: My apologies for imprecise review comment location. See the first <p> in the "sotd" section beginning in line 114. Consider changing it to something like the following: (It might also be a good time to put this list of substantive changes in an ordered list, since it has grown quite long and lacks consistent parallel sentence structure otherwise.)
``
      <p>On top of editorial updates, substantives changes since publication as a W3C Recommendation in <a href="https://www.w3.org/TR/2016/REC-media-source-20161117/">November 2016</a> are the addition of a {{SourceBuffer/changeType()}} method to switch codecs, the possibility to create and use {{MediaSource}} objects off the main thread in dedicated workers, the removal of the <code>createObjectURL()</code> extension to the {{URL}} object following its integration in the File API [[FILEAPI]], and the addition of {{ManagedMediaSource}}, ManagedSourceBuffer}}, and {{BufferedChangeEvent}} interfaces supporting power-efficient streaming and active buffered media cleanup by the user agent. For a full list of changes done since the previous version, see the <a href="https://github.com/w3c/media-source/commits/main">commits</a>.</p>
``

media-source-respec.html
```diff
@@ -2719,6 +2711,280 @@ <h3>Event Summary</h3>
...
     </section>
 
+    <section data-cite="mimesniff" data-dfn-for="ManagedMediaSource">
+      <h2><dfn>ManagedMediaSource</dfn> interface</h2>
+      <p>
+        A {{ManagedMediaSource}} is a {{MediaSource}} that actively manages its
+        memory content. Unlike a {{MediaSource}}, the [=user agent=] can evict content
+        through the [=ManagedMediaSource/memory
+        cleanup=] algorithm from its {{MediaSource/sourceBuffers}} (populated with {{ManagedSourceBuffer}})
+        for any reason.
```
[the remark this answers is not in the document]
padenot: It's not unusual to only say what should happen in spec, and not what should not happen (because most thing shouldn't happen). But this would be nice in a non-normative note (this section is normative).

padenot (approved):
Marking this as OK from the Mozilla perspective, keeping in mind a number of comments we had are going to be handled in follow-up.

Please don't hesitate to flag one of for review on those if we can help!

media-source-respec.html
```diff
@@ -2720,6 +2712,280 @@ <h3>Event Summary</h3>
...
+        <li>
+          <dl class="switch">
+            <dt>
+              If [=this=] is not in [=this=]'s {{ManagedMediaSource}} parent
+              {{MediaSource/activeSourceBuffers}}:
+            </dt>
+            <dd>
+              <ol>
+                <li>
+                  Run the [=coded frame removal=] algorithm with 0 and positive infinity equal to the removal range start and end timestamp respectively, and abort these steps.
```
chrisn, line 2963: A minor not, for consistency with how this is phrased elsewhere in the spec I'd suggest phrasing as "run the coded frame removal algorithm with start and end equal to 0 and positive infinity respectively.

media-source-respec.html
```diff
@@ -2719,6 +2711,280 @@ <h3>Event Summary</h3>
...
+                </li>
+              </ol>
+            </dd>
+          </dl>
+        </li>
+        <li>
+          Let |removal ranges:normalized TimeRanges| equal a list of presentation
+          time ranges that can be evicted from the presentation to ensure
+          uninterrupted playback from {{HTMLMediaElement.currentTime}} until such
+          presentation could be fetched again.
```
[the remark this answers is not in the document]
jyavenard: Ultimately, everything will be fetched from somewhere. So you may not fetch directly the content, but you will be fetching that other source that has content. So I don't think such distinction matters.

media-source-respec.html
```diff
@@ -2720,6 +2712,280 @@ <h3>Event Summary</h3>
...
+                  Run the [=coded frame removal=] algorithm with 0 and positive infinity equal to the removal range start and end timestamp respectively, and abort these steps.
+                </li>
+              </ol>
+            </dd>
+          </dl>
+        </li>
+        <li>
+          Let |removal ranges:normalized TimeRanges| equal a list of presentation
+          time ranges that can be evicted from the presentation to ensure
+          uninterrupted playback from {{HTMLMediaElement.currentTime}} until such
```
marcoscaceres, line 2972:
```suggestion
          uninterrupted playback from {{HTMLMediaElement/currentTime}} until such
```

media-source-respec.html
```diff
@@ -2719,6 +2711,280 @@ <h3>Event Summary</h3>
...
+          presentation could be fetched again.
+          <p class="note">
+            Implementations can use different strategies for selecting |removal ranges|
+            so web applications shouldn't depend on a specific behavior. The web
```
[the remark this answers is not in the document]
jyavenard: I had never seen that explainer/proposal before. It's unfortunate. The aim of this blabber was to make it as vague as possible as to how it could be done and preserve the intent. 

If the aim is to make that explainer official, and have it referenced here, it needs to be made official first.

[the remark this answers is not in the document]
marcoscaceres: Ok, yeah, let's consider bring the WICG proposal over as it's in scope of this work. We can probably just add it to the spec, but we would need to check what Firefox does and also what WebKit does (and how it aligns with Chrome). 

We could test for this maybe?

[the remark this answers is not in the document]
marcoscaceres: Conclusion is really that authors can really depend on this... a that stuff gets evicted for (implementation specific or hardware limits) "reasons".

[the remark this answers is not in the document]
jyavenard: To add: Firefox and Safari eviction policy are very similar:
- Evict up to the last keyframe before current time
- Evict any data in the future not contiguous in the buffer ahead of currentTime

media-source-respec.html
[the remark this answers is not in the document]
marcoscaceres: Just to add a bit more formality: it's all WHATWG Fetch spec under the hood for all fetches on the web platform, so we should make sure we are linking to the fetch spec when we talk about fetching.

media-source-respec.html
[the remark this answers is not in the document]
marcoscaceres: Filed bug for followup #341 ... treating this as resolved for now.

media-source-respec.html
jyavenard: There is no intention to separate the two, and I don't believe players would care. In fact I believe they should not differentiate: particularly because when appending data, it was possible for data to be evicted first (through the prepare append algorithm). The idea that the player could somehow assume that when they appended data nothing else was removed was bad, and has been the source of a lot of error in the wild.

How they would deal with content added and/or removed would pretty much be independent, then to check that there's data where they expect it to be.

media-source-respec.html
```diff
@@ -2696,6 +2688,280 @@ <h3>Event Summary</h3>
...
+    <section data-cite="mimesniff" data-dfn-for="ManagedMediaSource">
+      <h2><dfn>ManagedMediaSource</dfn> interface</h2>
+      <p>
+        A {{ManagedMediaSource}} is a {{MediaSource}} that actively manages its
+        memory content. Unlike a {{MediaSource}}, the [=user agent=] can evict content
+        through the [=ManagedMediaSource/memory
+        cleanup=] algorithm from its {{MediaSource/sourceBuffers}} (populated with {{ManagedSourceBuffer}})
+        for any reason.
+      </p>
+
```
marcoscaceres, line 2700:
```suggestion
<aside class="note" title="Eviction reasons">
<p>
  Reasons that the user agent might evict content are implementation specific and can include, but are not limited to, memory and/or hardware limitations, change in environmental conditions, and so on. Developers shouldn't make assumptions as to why, how, or when a user agent might evict content. Instead, developers need to write scripts with the assumption that content is constantly and randomly being evicted to avoid stalled video playback (i.e., code defensibly!). 
</p>
<aside>
```

media-source-respec.html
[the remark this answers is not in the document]
marcoscaceres: I wrote a note below to address this https://github.com/w3c/media-source/pull/329/files#r1414960629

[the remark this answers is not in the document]
marcoscaceres: See aside below...

media-source-respec.html
```diff
@@ -2696,6 +2688,284 @@ <h3>Event Summary</h3>
...
+      <p>
+        A {{ManagedMediaSource}} is a {{MediaSource}} that actively manages its
+        memory content. Unlike a {{MediaSource}}, the [=user agent=] can evict content
+        through the [=ManagedMediaSource/memory
+        cleanup=] algorithm from its {{MediaSource/sourceBuffers}} (populated with {{ManagedSourceBuffer}})
+        for any reason.
+      </p>
+<aside class="note" title="Eviction reasons">
+<p>
+  Reasons that the user agent might evict content are implementation specific and can include, but are not limited to, memory and/or hardware limitations, change in environmental conditions, and so on. Developers shouldn't make assumptions as to why, how, or when a user agent might evict content. Instead, developers need to write scripts with the assumption that content is constantly and randomly being evicted to avoid stalled video playback (i.e., code defensibly!).
```
marcoscaceres, line 2702:
```suggestion
  Reasons that the user agent might evict content are implementation specific and can include, but are not limited to, memory and/or hardware limitations, change in environmental conditions, and so on. Developers shouldn't make assumptions as to why, how, or when a user agent might evict content. Instead, developers need to write scripts with the assumption that content is constantly and randomly being evicted to avoid stalled video playback (i.e., code defensibly and listen for the `bufferedchangeevent`!). 
```

media-source-respec.html
marcoscaceres: Further, making the distinction would reveal implementation details that can't be relied on, as they can change from one browser version to another or on different hardware. 

So, I don't think we say anything further here.

media-source-respec.html
marcoscaceres: The attribute is not queried, but represents the state of the coalesced changes that occurred during the last completed operation (between updatestart and updateend... and whatever was removed if there was memory pressure, which can't be distinguished anyway).

media-source-respec.html
```diff
@@ -2696,6 +2688,284 @@ <h3>Event Summary</h3>
...
+          TimeRanges removedRanges;
+        };
+      </pre>
+      <h3>Attributes</h3>
+      <dl>
+        <dt>
+          <dfn>addedRanges</dfn>
+        </dt>
+        <dd>
+          The time ranges added during the last run of the [=coded frame processing=] algorithm.
```
jyavenard, line 2837:
```suggestion
          The time ranges added between the last `updatestart` and `updateeend` events (which would have occurred during the last run of the [=coded frame processing=] algorithm).
```

media-source-respec.html
```diff
@@ -2696,6 +2688,284 @@ <h3>Event Summary</h3>
...
+        </dd>
+        <dt>
+          <dfn>removedRanges</dfn>
+        </dt>
+        <dd>
+          The time ranges removed during the last run of the [=coded frame removal=] or [=coded frame eviction=] algorithm or if the
```
jyavenard, line 2843:
```suggestion
          The time ranges removed between the last `updatestart` and `updateend` events (which would have occurred during the last run of the [=coded frame removal=] or [=coded frame eviction=] algorithm or if the
```

media-source-respec.html
```diff
@@ -2696,6 +2688,284 @@ <h3>Event Summary</h3>
...
+          user agent evicted content in response to a [=ManagedSourceBuffer/memory cleanup=].
```
jyavenard, line 2844:
```suggestion
          user agent evicted content in response to a [=ManagedSourceBuffer/memory cleanup=]).
```

media-source-respec.html
```diff
@@ -2696,6 +2688,285 @@ <h3>Event Summary</h3>
...
+        <li>
+          <dl class="switch">
+            <dt>
+              If [=this=] is not in [=this=]'s {{ManagedMediaSource}} parent
+              {{MediaSource/activeSourceBuffers}}:
+            </dt>
+            <dd>
+              <ol>
+                <li>
+                  Run the [=coded frame removal=] algorithm with 0 and positive infinity equal to the removal range start and end timestamp respectively, and abort these steps.
```
marcoscaceres, line 2944:
```suggestion
                  Run the [=coded frame removal=] algorithm with start set to 0, end set to positive infinity, and abort these steps.
```

media-source-respec.html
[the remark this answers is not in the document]
marcoscaceres: Added @wolenetz suggested text to the SOTD.

marcoscaceres:
Ok, I think we have addressed all outstanding feedback. Thanks again @wolenetz for the detailed review. 

Can you check you are satisfied with the responses and give a tick ✅ if all looks good. 
Please feel free to make any further suggesting directly in the pull request.

media-source-respec.html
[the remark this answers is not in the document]
chrisn: I think the point here is that the content could be coming from something other than Fetch, e.g., RTCDataChannel.

media-source-respec.html
```diff
@@ -2696,6 +2691,285 @@ <h3>Event Summary</h3>
...
+                </li>
+              </ol>
+            </dd>
+          </dl>
+        </li>
+        <li>
+          Let |removal ranges:normalized TimeRanges| equal a list of presentation
+          time ranges that can be evicted from the presentation to ensure
+          uninterrupted playback from {{HTMLMediaElement/currentTime}} until such
+          presentation could be fetched again.
```
jyavenard, line 2957:
```suggestion
          presentation could be retrieved again.
```

media-source-respec.html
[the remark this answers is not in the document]
jyavenard: I changed the word "fetched" for "retrieved"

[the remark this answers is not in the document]
chrisn: Sounds good, thanks!

chrisn:
I have just one question: does ManagedMediaSource work with [MediaSourceHandle](https://w3c.github.io/media-source/#mediasourcehandle), to enable use of ManagedMediaSource in a Worker, and are there test cases for this?

jyavenard:
> I have just one question: does ManagedMediaSource work with [MediaSourceHandle](https://w3c.github.io/media-source/#mediasourcehandle), to enable use of ManagedMediaSource in a Worker, and are there test cases for this?

There's nothing specific about ManagedMediaSource when it comes to working in a Worker. All the apply for MediaSource should apply.

webkit doesn't support that feature at present.

dzianis-dashkevich:
@jyavenard 

> There's nothing specific about ManagedMediaSource when it comes to working in a Worker. All the apply for MediaSource should apply.

It seems like MSE in workers should be available in Safari 18:

https://developer.apple.com/documentation/safari-release-notes/safari-18-release-notes#Media

> Added support for MSE in workers. (123052315)

`ManagedMediaSource.canConstructInDedicatedWorker` returns true, but this code does not work as expected on Desktop and iOS for Safari 18. `ManagedMediaSource` never fires `sourceopen` event:

```js
const workerString = `
let ms;
let mms;
if ('MediaSource' in self) {
  ms = new MediaSource();

  ms.onsourceopen = () => console.log('ms-open');
  ms.onsourceended = () => console.log('ms-ended');
  ms.onsourceclose = () => console.log('ms-close');

  postMessage({ type: 'ms-handle', handle: ms.handle }, [ms.handle]);
}

if ('ManagedMediaSource' in self) {
  mms = new ManagedMediaSource();

  mms.onsourceopen = () => console.log('mms-open');
  mms.onsourceended = () => console.log('mms-ended');
  mms.onsourceclose = () => console.log('mms-close');

  postMessage({ type: 'mms-handle', handle: mms.handle }, [mms.handle]);
}
`;

const blob = new Blob([workerString], { type: 'application/javascript' });
const workerUrl = URL.createObjectURL(blob);
const worker = new Worker(workerUrl);

const videoElement1 = document.createElement('video');
const videoElement2 = document.createElement('video');

worker.onmessage = (e) => {
  console.log('received message from worker: ', e.data.type);

  if (e.data.type === 'ms-handle') {
    videoElement1.srcObject = e.data.handle;
  }

  if (e.data.type === 'mms-handle') {
    videoElement2.srcObject = e.data.handle;
  }
};

```

Here is the output for Safari 18 Desktop:

```shell
[Log] received message from worker:  – "ms-handle" (main.js, line 33)
[Log] received message from worker:  – "mms-handle" (main.js, line 33)
[Log] ms-open (8f6b0cc7-ba5b-4825-b632-6a66e24e3083, line 7)
```

Here is the output for Safari 18 iOS:

```shell
[Log] received message from worker:  – "mms-handle" (main.js, line 33)
```

jyavenard:
This repo isn't for discussing implementation issues and should be restricted to specs talk only.

In WebKit, for MSE and MMS to move from closed to open readyState, you must provide an alternative source that supports AirPlay, or you must explicitly disable remote playback
see https://jyavenard.github.io/htmltests/tests/ManagedMediaSource/bipbop.html

MSE isn't supported on iPhone, only MMS

PR #337, opened 2023-11-10
Editorial: let ReSpec handle id assignment

marcoscaceres:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/media-source/pull/337.html" title="Last updated on Dec 4, 2023, 4:50 AM UTC (adbe06c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/media-source/337/6f69e59...adbe06c.html" title="Last updated on Dec 4, 2023, 4:50 AM UTC (adbe06c)">Diff</a>

### DIFF
--- a/media-source-respec.html
+++ b/media-source-respec.html
@@ -145,49 +145,49 @@
         <h3>Definitions</h3>
 
         <dl>
-          <dt><dfn id="active-track-buffers">Active Track Buffers</dfn></dt>
+          <dt><dfn>Active Track Buffers</dfn></dt>
           <dd><p>The [=track buffers=] that provide [=coded frames=] for the {{AudioTrack/enabled}}
               {{HTMLMediaElement/audioTracks}}, the {{VideoTrack/selected}} {{HTMLMediaElement/videoTracks}}, and the
               <a def-id="texttrackmode-showing"></a> or <a def-id="texttrackmode-hidden"></a> {{HTMLMediaElement/textTracks}}. All these tracks are associated with
             <a>SourceBuffer</a> objects in the {{MediaSource/activeSourceBuffers}} list.</p>
           </dd>
 
-          <dt><dfn id="append-window">Append Window</dfn></dt>
+          <dt><dfn>Append Window</dfn></dt>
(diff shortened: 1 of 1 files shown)


tidoust (changes requested):
This update changes the effective ID of exported definitions. For instance, it will change `init-segment` to `dfn-initialization-segment`.

The problem is that byte stream format specs reference these exported terms, e.g., see links in the [Initialization Segments section](https://w3c.github.io/mse-byte-stream-format-mp2t/#mpeg2ts-init-segments) in MPEG-2 TS Byte Stream Format.

These specs have the same issue as MSE: they still use custom JavaScript to manage cross-references. They would certainly benefit from being updated as well, but since we don't really have plans to update the published byte stream formats specs any time soon, I would suggest to keep the explicit ID for terms that are exported.

No problem for non-exported terms.

tidoust:
> MPEG-2 TS Byte Stream Format is a tiny note though. Couldn't we (by which I mean I) update it and just put that on auto publish too? These are just spec heath changes so shouldn't impact anything... we could even keep the same date for the Note if we wanted to and no one would notice the difference.

We sure can and I'm happy to help. Note that we need to update the [5 byte stream format specifications](https://github.com/w3c/media-source#byte-stream-format-specifications), which all expect the `init-segment` term, I think. They're all very small too though, so totally manageable.

tidoust:
@marcoscaceres I prepared a refresh of the ISO BMFF Byte Stream Format spec to see what that would involve, see https://github.com/w3c/mse-byte-stream-format-isobmff/pull/12. If that looks fine, happy to do the same for the other specs (and then we can setup auto publish).

marcoscaceres:
That would be amazing! Thanks for handling the other doc!

marcoscaceres:
@tidoust, please merge this when you are ready in coordination with the other PR when ready. Thanks again!

tidoust:
> @tidoust, please merge this when you are ready in coordination with the other PR when ready. Thanks again!

Byte stream format specs and registry have now been updated not to rely on the custom `media-source.js` logic anymore. Merging accordingly.

Issue #347, opened 2024-02-16
Normative or informative definitions


chrisn:
All of the byte stream format specs (such as [this one](https://w3c.github.io/mse-byte-stream-format-isobmff/)) have ReSpec warnings because they refer to informative definitions in the MSE spec, e.g., "media segment", "initialization segment".

Should we make these definitions normative - e.g., by moving the current section 1.1 Definitions to a new section 2, so it's outside the informative section 1 Introduction? Some of the definitions include normative requirements on implementations.

tidoust:
+1 to make the definitions normative. On top to the byte stream format specs, the MSE spec itself references these informative definitions from within normative sections and algorithms, which isn't ideal either.

chrisn:
Resolved in #352

PR #352, opened 2024-03-20
Move Definitions section from Introduction

chrisn:
See #347.

### DIFF
--- a/media-source-respec.html
+++ b/media-source-respec.html
@@ -154,288 +154,288 @@
           </li>
         </ul>
       </section>
-      <section id="definitions">
-        <h3>
-          Definitions
-        </h3>
-        <dl>
-          <dt>
-            <dfn>Active Track Buffers</dfn>
-          </dt>
-          <dd>
-            <p>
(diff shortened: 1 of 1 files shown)


media-source-respec.html
```diff
@@ -154,288 +154,288 @@ <h3>
...
-            <p>
-              A Track ID is a byte stream format specific identifier that marks sections of the
-              byte stream as being part of a specific track. The Track ID in a [=track
-              description=] identifies which sections of a [=media segment=] belong to that track.
-            </p>
-          </dd>
-        </dl>
-      </section>
+    </section>
+    <section id="definitions">
```
tidoust, line 158: Seems a good opportunity to drop the ID. ReSpec generates the same ID, so that does not change anything.

```suggestion
    <section>
```

tidoust (approved):

media-source-respec.html
chrisn: Thanks, done.

marcoscaceres (approved):
lgtm

Issue #312, opened 2022-05-06
Editorial: Add image description text for the audio splice rendering image (in current WD section 4.5.12)


wolenetz:
This issue tracks the following portion of #307:

> §3.5.12 - The image representing the algorithm lacks a description. It's acknowledged that the image is itself a graphical representation of the algorithm described above it, but looks like it's a particularly concrete example, and it has specific features that should be described to help people interpret it. Another way to phrase this would be that the image isn't what I had in mind when reading the algorithm (though that may be due to lack of experience). For example, there are vertical lines and gradients at specific times in the image. It seems like a separate description may be warranted.

Note that this section, in the current WD, is now 4.5.12, and may change. It is the image example of the audio splice rendering algorithm.

Note also that the reference to external descriptive text (like the pipeline model image description) may need to refer directly to the ED version since it is not auto-published. See also #308 and #311, and consult @tidoust for best way to add this so both ED and WD get it.

chrisn:
The image is now in section 5.5.12.

PR #366, opened 2025-06-24
Add image description text for the audio splice rendering image

chrisn:
See #312

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/media-source/pull/366.html" title="Last updated on Nov 4, 2025, 11:30 PM UTC (0946914)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/media-source/366/612d3f5...0946914.html" title="Last updated on Nov 4, 2025, 11:30 PM UTC (0946914)">Diff</a>


marcoscaceres:
Having multiple files seems sub optimal, as the description will likely fall out of sync. Can we instead just throw the relevant information into a `<details>` element? Also, the information should be written to be suitable for all audiences, not specifically for someone using assistive technology (it seems somewhat excessive in its description).

media-source-respec.html
```diff
@@ -125,7 +125,7 @@ <h2>
       </p>
       <figure>
         <a href=
-        'https://w3c.github.io/media-source/pipeline_model_description.html#pipelinedesc'><picture><img src="pipeline_model.svg"
```
tidoust, line 128: FWIW, the spec-prod action used not to support automatic publication of local HTML assets linked from a spec, and the absolute URL was thus needed (the file does not exist under www.w3.org/TR yet). But that's no longer the case, that feature got added to spec-prod 3 years ago (and I of course forgot to update the URL in MSE accordingly!).

tidoust (approved):
That looks good to me, thanks for addressing the issue!

tidoust:
Hadn't seen @marcoscaceres's comment before I reviewed. This PR has the advantage of using a consistent scheme and approach for both images in the spec, but I don't disagree that the description (in both cases) seems excessive and would better be integrated within the main spec. Up to you :)

marcoscaceres (changes requested):
Let's move what we can into the main document. Having these kind of linked descriptions don't make a lot of sense and can quickly get out of sync.

chrisn:
I've removed the additional page, and add a `<details>` element. I shortened the text by just keeping the first paragraph. It's is enough to describe what's going on, I think.

@marcoscaceres @tidoust, please review. Thanks!

tidoust (approved):
Looks good. I guess we should also apply the same logic to the pipeline model diagram afterwards.

chrisn:
Sounds good, yes. I've adjusted the pipeline model description in this PR.

media-source-respec.html
chrisn: The PR now adds this description to the spec and removes the separate linked page.

media-source-respec.html
```diff
@@ -124,13 +124,70 @@ <h2>
...
+
+        <h3 id='sourcebuffer3'>Flowing down from SourceBuffer 3</h3>
+
+        <p>Flowing down from SourceBuffer 3 is a rose triangle with an arrow flowing to
+        a process indication labeled Track Buffer.<p>
+
+        <p>This Track Buffer then flows to a green box labeled Audio Decoder, which
+        then flows through a closed switch, which then also continues to the same
+        summing junction previously mentioned in SourceBuffer 1, which connects to the
+        Audio Device.</p>
```
tidoust: I'm thinking it could be useful to use variables to track boxes in this description. See updated text. Up to you.

Note two typo fixes in there:
 s/desination/destination
s/ouptut/output

```suggestion
        <p>The image shows a flowchart that reads from a top down perspective.
        At the top is a container
        marked |MediaSource|, with three arrows flowing down to containers each marked
        |SourceBuffer|. (Not indicated, but for sake of clarity in this description
        referred to here-after as |SourceBuffer 1|, shown with a blue background;
        |SourceBuffer 2|, shown with a mauve background; and |SourceBuffer 3|, shown
        with a rose background; as described from Left to Right).</p>

        <p>Below these three |SourceBuffer| containers is a dashed line, with an indication
        that the top half (|MediaSource| plus the three |SourceBuffer| containers) are taken
        from the Media Source API, while the details to follow are taken from the
        {{HTMLMediaElement}}.<p>

        <h3 id='sourcebuffer1'>Flowing down from SourceBuffer 1</h3>

        <p>Flowing down from |SourceBuffer 1| is a blue triangle with three arrows pointing
        to three process indications, each labeled |Track Buffer|.</p>

        <p>The first |Track Buffer| from the left then flows to a red box labeled |Video
        Decoder|, which then flows to a switch that selects between one of three inputs,
        and the output of the switch connects to an output destination, labeled |Video Tag
        Display Region|. The switch indicates that one |Video Decoder| output (from
        |SourceBuffer 2|) is selected to flow into the |Video Tag Display Region|.</p>

        <p>The second |Track Buffer| flows to a green box labeled |Audio Decoder|, which
        then follows to an indicator of an open switch, which then continues to a
        summing junction indicator. The output of the summing junction connects to
        a green output destination, labeled |Audio Device|.</p>

        <p>The third |Track Buffer| also flows to a green box labeled |Audio Decoder|,
        which also flows to a switch, this time however indicated closed, which then
        also continues to the same summing junction as the second |Track Buffer|,
        which connects to the |Audio Device|.</p>

        <h3 id='sourcebuffer2'>Flowing down from SourceBuffer 2</h3>

        <p>Flowing down from |SourceBuffer 2| is a mauve triangle with an arrow flowing
        to a process indication labeled |Track Buffer|.<p>

        <p>This |Track Buffer| then flows to a red box labeled |Video Decoder|, which
        then flows to the same switch that feeds to the |Video Tag Display Region|
        previously mentioned in |SourceBuffer 1|. The switch is positioned so that
        this |Video Decoder| output is connected to the |Video Tag Display Region|.</p>

        <h3 id='sourcebuffer3'>Flowing down from SourceBuffer 3</h3>

        <p>Flowing down from |SourceBuffer 3| is a rose triangle with an arrow flowing to
        a process indication labeled |Track Buffer|.<p>

        <p>This |Track Buffer| then flows to a green box labeled |Audio Decoder|, which
        then flows through a closed switch, which then also continues to the same
        summing junction previously mentioned in |SourceBuffer 1|, which connects to the
        |Audio Device|.</p>
```

media-source-respec.html
```diff
@@ -124,13 +124,70 @@ <h2>
...
-        <a href=
-        'https://w3c.github.io/media-source/pipeline_model_description.html#pipelinedesc'><picture><img src="pipeline_model.svg"
-        alt="Media Source Pipeline Model Diagram"></picture></a>
+        <picture><img src="pipeline_model.svg"
+        alt="Media Source Pipeline Model Diagram"></picture>
         <figcaption>
           Media Source Pipeline Model Diagram
         </figcaption>
       </figure>
+      <details class="note">
```
marcoscaceres: I think the details should be moved inside the `<figcaption>` (or ideally removed entirely).

media-source-respec.html
```diff
@@ -124,13 +124,70 @@ <h2>
...
+        referred to here-after as |SourceBuffer 1|, shown with a blue background;
+        |SourceBuffer 2|, shown with a mauve background; and |SourceBuffer 3|, shown
+        with a rose background; as described from Left to Right).</p>
+
+        <p>Below these three |SourceBuffer| containers is a dashed line, with an indication
+        that the top half (|MediaSource| plus the three |SourceBuffer| containers) are taken
+        from the Media Source API, while the details to follow are taken from the
+        {{HTMLMediaElement}}.<p>
+
+        <h3 id='sourcebuffer1'>Flowing down from SourceBuffer 1</h3>
```
marcoscaceres:
```suggestion
        <h3 id='sourcebuffer1' class="notoc">Flowing down from SourceBuffer 1</h3>
```

media-source-respec.html
```diff
@@ -124,13 +124,70 @@ <h2>
...
+        then follows to an indicator of an open switch, which then continues to a
+        summing junction indicator. The output of the summing junction connects to
+        a green output destination, labeled |Audio Device|.</p>
+
+        <p>The third |Track Buffer| also flows to a green box labeled |Audio Decoder|,
+        which also flows to a switch, this time however indicated closed, which then
+        also continues to the same summing junction as the second |Track Buffer|,
+        which connects to the |Audio Device|.</p>
+
+        <h3 id='sourcebuffer2'>Flowing down from SourceBuffer 2</h3>
```
marcoscaceres:
```suggestion
        <h3 id="sourcebuffer2" class="notoc">Flowing down from SourceBuffer 2</h3>
```

media-source-respec.html
```diff
@@ -124,13 +124,70 @@ <h2>
...
+
+        <p>Flowing down from |SourceBuffer 2| is a mauve triangle with an arrow flowing
+        to a process indication labeled |Track Buffer|.<p>
+
+        <p>This |Track Buffer| then flows to a red box labeled |Video Decoder|, which
+        then flows to the same switch that feeds to the |Video Tag Display Region|
+        previously mentioned in |SourceBuffer 1|. The switch is positioned so that
+        this |Video Decoder| output is connected to the |Video Tag Display Region|.</p>
+
+        <h3 id='sourcebuffer3'>Flowing down from SourceBuffer 3</h3>
```
marcoscaceres:
```suggestion
        <h3 id="sourcebuffer3" class="notoc">Flowing down from SourceBuffer 3</h3>
```

media-source-respec.html
```diff
@@ -124,13 +124,70 @@ <h2>
...
+
+        <p>Flowing down from |SourceBuffer 3| is a rose triangle with an arrow flowing to
+        a process indication labeled |Track Buffer|.<p>
+
+        <p>This |Track Buffer| then flows to a green box labeled |Audio Decoder|, which
+        then flows through a closed switch, which then also continues to the same
+        summing junction previously mentioned in |SourceBuffer 1|, which connects to the
+        |Audio Device|.</p>
+
+        <p style='text-align: right;'>With thanks to <a href='https://lists.w3.org/Archives/Public/public-apa/2016Jan/0014.html'>J. Foliot, Deque Systems</a></p>
```
marcoscaceres: Should be moved to acknowledgements section instead.

marcoscaceres:
I added `<desc>` to the SVG file, and got chatty to write an appropriate summary based on the current PR. 

I still think this is excessive and we shouldn't add the `<details>` element. The SVGs' new `<desc>` should suffice and we should extend that description if need be.

media-source-respec.html
```diff
@@ -4293,7 +4350,23 @@ <h4>
...
+            </figure>
+            <details>
+              <summary>Image description</summary>
+              <p>
+                The image depicts the process of audio cross-fading between two sets of coded
+                frames over a duration of 5 milliseconds. The audio to be faded-out decreases
+                in volume from maximum to zero over a 5 millisecond period, while the audio
+                to be faded-in increases in volume from zero to maximum over the same period.
+              </p>
+            </details>
```
marcoscaceres:
```suggestion
            </p>
            <figure>
              <picture><img src="audio_splice.png"
              alt="Audio Splice Rendering diagram"></picture>
              <figcaption>
                The image depicts the process of audio cross-fading between two sets of coded
                frames over a duration of 5 milliseconds. The audio to be faded-out decreases
                in volume from maximum to zero over a 5 millisecond period, while the audio
                to be faded-in increases in volume from zero to maximum over the same period.
              </figcaption>
            </figure>
```

media-source-respec.html
```diff
@@ -124,13 +124,70 @@ <h2>
         extensions are available in the byte stream format registry [[MSE-REGISTRY]].
       </p>
       <figure>
-        <a href=
-        'https://w3c.github.io/media-source/pipeline_model_description.html#pipelinedesc'><picture><img src="pipeline_model.svg"
-        alt="Media Source Pipeline Model Diagram"></picture></a>
+        <picture><img src="pipeline_model.svg"
+        alt="Media Source Pipeline Model Diagram"></picture>
         <figcaption>
           Media Source Pipeline Model Diagram
```
marcoscaceres:
```suggestion
            Media Source pipeline diagram showing how a MediaSource provides one or more SourceBuffers
            that feed audio and video decoders through a media element. The decoded data is rendered
            as synchronized audio and video output.
```

media-source-respec.html
chrisn: Done

marcoscaceres (changes requested):
I'm still very much opposed to this change, on accessibility grounds. Particularly this is violation of: 

 * [Avoid Too Much Content](https://www.w3.org/WAI/WCAG2/supplemental/patterns/o5p03-manageable-quantity/): "Too much information or instruction can be just as much of a hindrance as too little. The goal is to make certain that enough information is provided for the user to accomplish the task without undue confusion or navigation."
 * Particularly, having the headings in there was super confusing. 

The correct way to handle this is three fold:

 * The spec text itself should describe the what the flows are. If there is stuff in the detailed description that is not covered by the spec text, then we should fold that into there. 
 * I've expanded the figure caption text to describe the what is in the diagram. 
 * The SVG itself describes what is in the diagram using `desc`.

Issue #280, opened 2021-06-25
MSE-in-Workers: {Audio,Video,Text}Track{,List} IDL in HTML need additional DedicatedWorker in Exposed


wolenetz:
MSE-in-Workers feature issue is #175.

This issue tracks making sure the extended HTML AudioTrack, VideoTrack, TextTrack, and their *TrackList IDLs are all updated to have exposure to both Window and DedicatedWorker, not just Window.

In the initial spec draft of MSE-in-Workers, the assumption is made that such exposure is already included in HTML, but at the moment of filing this issue, it hasn't.

Per WebIDL https://heycam.github.io/webidl/#Exposed, it looks like HTML spec's original definitions of these IDL would need to include DedicatedWorker in their Exposed (otherwise, some significant rework or even removal of track/tracklist visibility in MSE-in-Worker would be necessary):

> If [Exposed] appears on a partial interface or partial namespace, then the partial’s own exposure set must be a subset of the exposure set of the partial’s original interface or namespace.

wolenetz:
Requesting this be discussed at Sept 14 Media WG meeting ([minutes](https://www.w3.org/2021/09/14-mediawg-minutes.html#t04)).

foolip:
I noticed this issue because HTML's and MSE's Web IDL put together are invalid right now, causing test failures in https://github.com/web-platform-tests/wpt/pull/29414.

I can't find an issue about this in HTML yet, would be great if someone could file an issue at https://github.com/whatwg/html/issues/new.

wolenetz:
I've reactivated the most closely related Chromium implementation bug (https://crbug.com/487288) to help track fixing this issue in implementation, too (currently behind the experimental AudioVideoTracks blink runtime-enabled feature).

jyavenard:
There's unfortunately much more DOM objects that need updating, in particular TextTrackCueList / TextTrackCue ; in TextTrack the method addCue and removeCue which have references on how they are to be rendered on the MediaElement (https://w3c.github.io/webvtt/#rules-for-updating-the-display-of-webvtt-text-tracks) with their respective CSS block boxes etc..
while those examples are in relation to WebVTT, there's nothing preventing the use of WebVTT cues in inbandTextTrack

None of which is applicable with DedicatedWorker

PR #376, opened 2026-07-08
TrackList and Tracks shouldn't be exposed in a DedicatedWorker.

### LINKED ISSUE #361: TrackList and Tracks shouldn't be exposed in a DedicatedWorker.
A SourceBuffer exposes all its tracks related objects in a Worker
https://w3c.github.io/media-source/#webidl-544711679
  readonly attribute [AudioTrackList](https://html.spec.whatwg.org/multipage/media.html#audiotracklist) [audioTracks](https://w3c.github.io/media-source/#dom-sourcebuffer-audiotracks);
  readonly  attribute [VideoTrackList](https://html.spec.whatwg.org/multipage/media.html#videotracklist) [videoTracks](https://w3c.github.io/media-source/#dom-sourcebuffer-videotracks);
  readonly  attribute [TextTrackList](https://html.spec.whatwg.org/multipage/media.html#texttracklist) [textTracks](https://w3c.github.io/media-source/#dom-sourcebuffer-texttracks);

Track objects are currently only exposed at the Window level 
An issue was open to expose them at the Worker level in #280 

As noted in #280 
There are unfortunately much more DOM objects that need updating, in particular TextTrackCueList / TextTrackCue ; in TextTrack the method addCue and removeCue which have references on how they are to be rendered on the MediaElement (https://w3c.github.io/webvtt/#rules-for-updating-the-display-of-webvtt-text-tracks) with their respective CSS block boxes etc..
while those examples are in relation to WebVTT, there's nothing preventing the use of WebVTT cues in inbandTextTrack

None of which is applicable with DedicatedWorker

Considering there are no User Agent other than webkit implementing Tracks objects in the SourceBuffer when in a DedicatedWorker

We think that those objects shouldn't be exposed in a Worker as they wouldn't be usable anyway.

jyavenard:
Make them Window-scope only and update various algorithms to remove creation of those objects in DedicatedWorker

Fixes #361

***
<a href="https://pr-preview.s3.amazonaws.com/jyavenard/media-source/pull/376.html" title="Last updated on Jul 8, 2026, 2:41 PM UTC (e4ca821)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/media-source/376/612d3f5...jyavenard:e4ca821.html" title="Last updated on Jul 8, 2026, 2:41 PM UTC (e4ca821)">Diff</a>

### DIFF
--- a/media-source-respec.html
+++ b/media-source-respec.html
@@ -765,46 +765,49 @@
             </li>
           </ol>
         </li>
-        <li>Let |SourceBuffer audioTracks list:AudioTrackList| equal the {{AudioTrackList}} object
-        returned by |sourceBuffer|.{{SourceBuffer/audioTracks}}.
-        </li>
-        <li>If the |SourceBuffer audioTracks list| is not empty, then run the following steps:
-          <ol>
-            <li>For each {{AudioTrack}} object in the |SourceBuffer audioTracks list|, run the
-            following steps:
+        <li>Run the appropriate steps from the following list:
+          <dl class="switch">
+            <dt>
+              If the {{MediaSource}} was constructed in a {{Window}}:
+            </dt>
+            <dd>
               <ol>
-                <li>Set the {{AudioTrack/sourceBuffer}} attribute on the {{AudioTrack}} object to
-                null.
+                <li>Let |SourceBuffer audioTracks list:AudioTrackList| equal the {{AudioTrackList}}
+                object returned by |sourceBuffer|.{{SourceBuffer/audioTracks}}.
                 </li>
-                <li>Remove the {{AudioTrack}} object from the |SourceBuffer audioTracks list|.
-                  <p class="note">
-                    This should trigger {{AudioTrackList}} [[HTML]] logic to [=queue a task=] to
-                    [=fire an event=] named [=AudioTrackList/removetrack=] using {{TrackEvent}}
-                    with the {{TrackEvent/track}} attribute initialized to the {{AudioTrack}}
-                    object, at the |SourceBuffer audioTracks list|. If the {{AudioTrack/enabled}}
-                    attribute on the {{AudioTrack}} object was true at the beginning of this
-                    removal step, then this should also trigger {{AudioTrackList}} [[HTML]] logic
-                    to [=queue a task=] to [=fire an event=] named [=AudioTrackList/change=] at the
-                    |SourceBuffer audioTracks list|.
-                  </p>
+                <li>Let |media element:HTMLMediaElement| be the media element that |sourceBuffer|'s
+                [=parent media source=] is attached to.
                 </li>
-                <li>Use the [=mirror if necessary=] algorithm to run the following steps in
-                {{Window}}, to remove the {{AudioTrack}} object (or instead, the {{Window}} mirror
-                of it if the {{MediaSource}} object was constructed in a
-                {{DedicatedWorkerGlobalScope}}) from the media element:
+                <li>Let |HTMLMediaElement audioTracks list:AudioTrackList| equal the
+                {{AudioTrackList}} object returned by the {{HTMLMediaElement/audioTracks}} attribute
+                on |media element|.
+                </li>
+                <li>For each |audio track:AudioTrack| in the |SourceBuffer audioTracks
+                list|, run the following steps:
                   <ol>
-                    <li>Let |HTMLMediaElement audioTracks list:AudioTrackList| equal the
-                    {{AudioTrackList}} object returned by the {{HTMLMediaElement/audioTracks}}
-                    attribute on the HTMLMediaElement.
+                    <li>Set the {{AudioTrack/sourceBuffer}} attribute on |audio track| to
+                    null.
                     </li>
-                    <li>Remove the {{AudioTrack}} object from the |HTMLMediaElement audioTracks
+                    <li>Remove |audio track| from the |SourceBuffer audioTracks list|.
+                      <p class="note">
+                        This should trigger {{AudioTrackList}} [[HTML]] logic to [=queue a task=] to
+                        [=fire an event=] named [=AudioTrackList/removetrack=] using {{TrackEvent}}
+                        with the {{TrackEvent/track}} attribute initialized to |audio track|, at the
+                        |SourceBuffer audioTracks list|. If the {{AudioTrack/enabled}}
+                        attribute on |audio track| was true at the beginning of this
+                        removal step, then this should also trigger {{AudioTrackList}} [[HTML]] logic
+                        to [=queue a task=] to [=fire an event=] named [=AudioTrackList/change=] at the
+                        |SourceBuffer audioTracks list|.
+                      </p>
+                    </li>
+                    <li>Remove |audio track| from the |HTMLMediaElement audioTracks
                     list|.
                       <p class="note">
                         This should trigger {{AudioTrackList}} [[HTML]] logic to [=queue a task=]
                         to [=fire an event=] named [=AudioTrackList/removetrack=] using
-                        {{TrackEvent}} with the {{TrackEvent/track}} attribute initialized to the
-                        {{AudioTrack}} object, at the |HTMLMediaElement audioTracks list|. If the
-                        {{AudioTrack/enabled}} attribute on the {{AudioTrack}} object was true at
+                        {{TrackEvent}} with the {{TrackEvent/track}} attribute initialized to
+                        |audio track|, at the |HTMLMediaElement audioTracks list|. If the
+                        {{AudioTrack/enabled}} attribute on |audio track| was true at
                         the beginning of this removal step, then this should also trigger
                         {{AudioTrackList}} [[HTML]] logic to [=queue a task=] to [=fire an event=]
                         named [=AudioTrackList/change=] at the |HTMLMediaElement audioTracks list|.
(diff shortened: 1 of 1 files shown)


media-source-respec.html
```diff
@@ -765,37 +765,37 @@ <h3>
...
-                  </p>
+                <li>Let |HTMLMediaElement audioTracks list:AudioTrackList| equal the
+                {{AudioTrackList}} object returned by the {{HTMLMediaElement/audioTracks}} attribute
+                on the HTMLMediaElement.
                 </li>
-                <li>Use the [=mirror if necessary=] algorithm to run the following steps in
-                {{Window}}, to remove the {{AudioTrack}} object (or instead, the {{Window}} mirror
-                of it if the {{MediaSource}} object was constructed in a
-                {{DedicatedWorkerGlobalScope}}) from the media element:
+                <li>For each {{AudioTrack}} object in the |SourceBuffer audioTracks list|, run the
```
marcoscaceres:
```suggestion
                <li>[=list/for each=] |track:AudioTrack| {{AudioTrack}} in the |SourceBuffer audioTracks list|, run the
```

media-source-respec.html
```diff
@@ -765,37 +765,37 @@ <h3>
...
+                following steps:
                   <ol>
-                    <li>Let |HTMLMediaElement audioTracks list:AudioTrackList| equal the
-                    {{AudioTrackList}} object returned by the {{HTMLMediaElement/audioTracks}}
-                    attribute on the HTMLMediaElement.
+                    <li>Set the {{AudioTrack/sourceBuffer}} attribute on the {{AudioTrack}} object to
```
marcoscaceres: Should this be done in a task in real time? or is it ok to be done here. It's suggesting that it's changing on main thread, but not sure.

media-source-respec.html
```diff
@@ -765,37 +765,37 @@ <h3>
...
+                    null.
+                    </li>
+                    <li>Remove the {{AudioTrack}} object from the |SourceBuffer audioTracks list|.
```
marcoscaceres:
```suggestion
                    <li>[=list/Remove=] the {{AudioTrack}} object from the |SourceBuffer audioTracks list|.
```

media-source-respec.html
```diff
@@ -765,37 +765,37 @@ <h3>
...
+                      <p class="note">
+                        This should trigger {{AudioTrackList}} [[HTML]] logic to [=queue a task=] to
```
marcoscaceres, line 793: This shouldn't be a "should" right? This is somehow wired to happen? 

```suggestion
                        This triggers {{AudioTrackList}} [[HTML]] logic to [=queue a task=] to
```

media-source-respec.html
```diff
@@ -765,37 +765,37 @@ <h3>
...
+                        [=fire an event=] named [=AudioTrackList/removetrack=] using {{TrackEvent}}
+                        with the {{TrackEvent/track}} attribute initialized to the {{AudioTrack}}
+                        object, at the |SourceBuffer audioTracks list|. If the {{AudioTrack/enabled}}
+                        attribute on the {{AudioTrack}} object was true at the beginning of this
+                        removal step, then this should also trigger {{AudioTrackList}} [[HTML]] logic
```
marcoscaceres, line 798:
```suggestion
                        removal step, then this will also trigger {{AudioTrackList}} [[HTML]] logic
```

media-source-respec.html
```diff
@@ -813,40 +813,71 @@ <h3>
...
+            <dt>
+              Otherwise:
+            </dt>
+            <dd>
+              Post an internal `remove track` message to {{MediaSource/[[port to main]]}} whose
+              implicit handler in {{Window}} runs the following steps:
+              <ol>
+                <li>Let |HTMLMediaElement audioTracks list:AudioTrackList| equal the
+                {{AudioTrackList}} object returned by the {{HTMLMediaElement/audioTracks}} attribute
+                on the HTMLMediaElement.
```
marcoscaceres:
```suggestion
                on the {{HTMLMediaElement}}.
```

media-source-respec.html
```diff
@@ -813,40 +813,71 @@ <h3>
...
+                </li>
+                <li>For each {{AudioTrack}} object in the |HTMLMediaElement audioTracks list| that
```
marcoscaceres:
```suggestion
                <li>For each |track:AudioTrack| {{AudioTrack}} in the |HTMLMediaElement audioTracks list| that
```

media-source-respec.html
```diff
@@ -813,40 +813,71 @@ <h3>
...
+                was created by |sourceBuffer|, run the following steps:
+                  <ol>
+                    <li>Remove the {{AudioTrack}} object from the |HTMLMediaElement audioTracks
```
marcoscaceres:
```suggestion
                    <li>{=list/Remove=] |track| from the |HTMLMediaElement audioTracks
```

media-source-respec.html
```diff
@@ -813,40 +813,71 @@ <h3>
...
+                    list|.
+                      <p class="note">
+                        This should trigger {{AudioTrackList}} [[HTML]] logic to [=queue a task=]
```
marcoscaceres, line 840: As above...

media-source-respec.html
```diff
@@ -813,40 +813,71 @@ <h3>
...
-                    with the {{TrackEvent/track}} attribute initialized to the {{VideoTrack}}
-                    object, at the |SourceBuffer videoTracks list|. If the {{VideoTrack/selected}}
-                    attribute on the {{VideoTrack}} object was true at the beginning of this
-                    removal step, then this should also trigger {{VideoTrackList}} [[HTML]] logic
-                    to [=queue a task=] to [=fire an event=] named [=VideoTrackList/change=] at the
-                    |SourceBuffer videoTracks list|.
-                  </p>
+                <li>Let |HTMLMediaElement videoTracks list:VideoTrackList| equal the
+                {{VideoTrackList}} object returned by the {{HTMLMediaElement/videoTracks}} attribute
+                on the HTMLMediaElement.
```
marcoscaceres:
```suggestion
                on the {{HTMLMediaElement}}.
```

marcoscaceres: Note...   on the {{HTMLMediaElement}} should really be a variable that is passed or gotten from somewhere.

media-source-respec.html
```diff
@@ -813,40 +813,71 @@ <h3>
...
                   <ol>
-                    <li>Let |HTMLMediaElement videoTracks list:VideoTrackList| equal the
-                    {{VideoTrackList}} object returned by the {{HTMLMediaElement/videoTracks}}
-                    attribute on the HTMLMediaElement.
+                    <li>Set the {{VideoTrack/sourceBuffer}} attribute on the {{VideoTrack}} object to
+                    null.
+                    </li>
+                    <li>Remove the {{VideoTrack}} object from the |SourceBuffer videoTracks list|.
+                      <p class="note">
+                        This should trigger {{VideoTrackList}} [[HTML]] logic to [=queue a task=] to
```
marcoscaceres, line 881: As above... also, are these notes needed... they seem redundant and overly specific... if there's an algorithm handling this, we should just point to that algorithm here.

marcoscaceres (commented):
This seems ok, but it's super duplicative... is there some way to get all this consolidated to be less repetitive?

marcoscaceres (commented):
Reviewed the exposure change against implementations. The Window-only direction matches what engines already do, so this reads as codifying existing behavior rather than changing it:

- WebKit (which ships MSE-in-Workers) already exposes `SourceBuffer` in `DedicatedWorker` while keeping `audioTracks`/`videoTracks`/`textTracks` as member-level `[Exposed=Window]`, which is the exact shape this PR specifies (`SourceBuffer.idl`); the `AudioTrack`/`VideoTrack`/`TextTrack` MSE partials are `[Exposed=Window]`.
- In Chromium, `SourceBuffer.audioTracks`/`videoTracks` sit behind the experimental `AudioVideoTracks` flag (off by default) and the base `AudioTrack` interface is `[Exposed=Window]`; there is no `SourceBuffer.textTracks`.
- In Gecko, `SourceBuffer.audioTracks` is commented out.

Moving the default enabled/selected decision into the `[[audio track added flag]]` / `[[video track added flag]]` slots looks correct: the worker `SourceBuffer` no longer exposes `audioTracks`, so the previous `audioTracks.length == 0` test would be unimplementable there, and a worker-side slot is the right replacement.

One housekeeping item: this looks like it also resolves #280 (which tracks the opposite direction, giving these IDLs Window+DedicatedWorker exposure). The inline issue boxes for #280 are removed here, so it would be worth closing or updating #280 and referencing it in the description alongside "Fixes #361".

As a follow-up, not for this PR: there do not appear to be any WPT tests asserting these attributes are absent in a `DedicatedWorkerGlobalScope`. An idlharness or exposure test would help lock the behavior in.

media-source-respec.html
```diff
@@ -2066,9 +2157,9 @@ <h2>
           readonly  attribute boolean updating;
           readonly  attribute TimeRanges buffered;
           attribute double timestampOffset;
-          readonly  attribute AudioTrackList audioTracks;
-          readonly  attribute VideoTrackList videoTracks;
-          readonly  attribute TextTrackList textTracks;
+          [Exposed=Window] readonly  attribute AudioTrackList audioTracks;
+          [Exposed=Window] readonly  attribute VideoTrackList videoTracks;
+          [Exposed=Window] readonly  attribute TextTrackList textTracks;
           attribute double appendWindowStart;
```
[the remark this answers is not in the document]
marcoscaceres: I'd suggest keeping the member-level `[Exposed=Window]` here rather than moving to an interface- or partial-level annotation. `SourceBuffer` itself needs to remain `[Exposed=(Window,DedicatedWorker)]`, so an interface-level `[Exposed=Window]` can't express "the interface is available in workers but these three members are not". Restricting a member to a subset of its interface's exposure set (Window is a subset of {Window, DedicatedWorker}) is valid Web IDL, and it is what WebKit ships today: `SourceBuffer` is `Exposed=(Window,DedicatedWorker)` with `audioTracks`/`videoTracks`/`textTracks` annotated `[Exposed=Window]` at the member level. The `AudioTrack`/`VideoTrack`/`TextTrack` partials use interface-level `[Exposed=Window]` because those interfaces are Window-only, which is a different case, so the mixed style here is intentional rather than inconsistent.

media-source-respec.html
jyavenard: yes, but this is re-using the existing text, I just moved it along. I didn't feel like re-writing existing part was in the scope of this PR. Those types of fix-up should be done in a follow-up PR IMHO

media-source-respec.html
```diff
@@ -864,44 +895,74 @@ <h3>
...
-          </ol>
-        </li>
-        <li>Let |SourceBuffer textTracks list:TextTrackList| equal the {{TextTrackList}} object
-        returned by |sourceBuffer|.{{SourceBuffer/textTracks}}.
+            </dd>
+            <dt>
+              Otherwise:
+            </dt>
+            <dd>
+              Post an internal `remove track` message to {{MediaSource/[[port to main]]}} whose
```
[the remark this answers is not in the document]
jyavenard: The original spec already reused a single message name across all three track types:
  - create track mirror — 3 occurrences (lines 3274/3385/3490 in the original), one each in the audio, video, and text
  creation handlers.
  - update track state — 2 occurrences, likewise not track-type-specific.
  
  My changes follow that exact pattern: create track is the (renamed) reuse of the original create track mirror across
  audio/video/text, and the new remove track mirrors it for the removal handlers. So making them track-type-specific
  (remove audio track, etc.) would deviate from the spec's established style, not improve consistency with it.

This PR isn't about changing existing text or patterns used throughout, but to do as the title states and only that. 

This can be taken as a follow-up

media-source-respec.html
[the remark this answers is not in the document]
jyavenard: I disagree, as @marcoscaceres wrote below

media-source-respec.html
jyavenard: there just moving the existing text, I'm only rewriting parts that require rewriting, not modifying existing wordings.

media-source-respec.html
jyavenard: |SourceBuffer audioTracks list| is an AudioTrackList — a WebIDL interface object, not an Infra list. The autolink won't resolve to anything meaningful; AudioTrackList doesn't expose an Infra list you can "for each" over.

media-source-respec.html
jyavenard: this is in the "If the MediaSource was constructed in a Window" section. By construction that
  means removeSourceBuffer() is running on the main thread, and the AudioTrack objects live on that same thread.  Setting |track|.sourceBuffer = null is a same-thread synchronous mutation

media-source-respec.html
jyavenard: also this is existing verbiage.

jyavenard: I've applied some light changes that take the spirit of your comment.

